PR #77, opened 2024-11-15
Trying to get deployment to work on Heroku for web front-end

tibetsprague:


apps/web/package.json
```diff
@@ -40,43 +46,28 @@
...
     "@deck.gl-community/layers": "^9.0.3",
-    "@deck.gl/core": "9.0.33",
-    "@deck.gl/extensions": "^9.0.33",
-    "@deck.gl/geo-layers": "9.0.33",
-    "@deck.gl/layers": "9.0.33",
-    "@deck.gl/mapbox": "^9.0.33",
-    "@deck.gl/mesh-layers": "9.0.33",
-    "@deck.gl/react": "9.0.33",
+    "@deck.gl/core": "9.0.35",
+    "@deck.gl/extensions": "^9.0.35",
```
thomasgwatson, line 52: did you want to remove the ^ here, just to make sure the deckgl versions issue doesn't return?

tibetsprague: sure

apps/web/package.json
```diff
@@ -6,21 +6,27 @@
...
+  },
+  "packageManager": "yarn@4.5.0",
   "browserslist": [
     "defaults"
   ],
   "proxy": "localhost:3001",
   "scripts": {
     "viteDev": "vite serve",
     "dev": "yarn start",
-    "start": "rm -rf ../../node_modules/.cache/babel-loader && node --preserve-symlinks scripts/start.js",
```
lorenjohnson, line 16: `rm -rf ../../node_modules/.cache/babel-loader` was needed in the past due to a bug in graqphql parsing which meant that changes to existing files would not be reflected even after a restart. Do you know if this cleared-up?

lorenjohnson, line 16: `rm -rf ../../node_modules/.cache/babel-loader` was needed in the past due to a bug in graqphql parsing which meant that changes to existing files would not be reflected even after a restart. Do you know if this cleared-up?

apps/backend/package.json
```diff
@@ -14,26 +14,33 @@
     ]
   },
   "main": "app.js",
+  "engines": {
+    "node": "^20",
+    "yarn": "^4.5.0"
+  },
+  "packageManager": "yarn@4.5.0",
   "scripts": {
-    "start": "yarn workspace @hylo/shared build:cjs && node --optimize_for_size --max_old_space_size=$MAX_OLD_SPACE_SIZE --gc_interval=100 app.js",
```
lorenjohnson, line 18: What was coming-up with this on-the-fly transpire on deploy?

package.json
```diff
@@ -7,12 +7,15 @@
...
+    "migrate": "yarn workspace backend migrate",
+    "worker": "yarn workspace backend worker",
     "mobile:dev": "yarn workspace mobile start",
     "mobile:pod-install": "yarn workspace mobile pod-install",
     "mobile:android": "yarn workspace mobile run android",
     "mobile:clean": "yarn workspace mobile run clean",
     "mobile:ios": "yarn workspace mobile run ios",
-    "reset": "rm -rf **/node_modules **/.yarn/cache **/yarn.lock && yarn install"
+    "reset": "rm -rf **/node_modules **/.yarn/cache **/yarn.lock && yarn install",
+    "heroku-postbuild": "yarn workspace web build && yarn workspace @hylo/shared build:cjs"
```
lorenjohnson, line 18: So this hook runs after the build on Heorku for both backend and web? If this works in deploy can we leave the on-the-fly @hylo/shared build for the `yarn workspace backend dev`?

lorenjohnson (approved):
See my comments/questions, the one about graphql and the babel cache clear on Web is the only one that feels blocking. Otherwise, just wanting more information on the others, but I don't want to hold up merging. I would like to understand the deploy issues better as I narrow in on getting our shared code picture clear for upcoming URQL merging.

apps/web/package.json
tibetsprague: i dont, but i will check

package.json
lorenjohnson: ok. let's do that and then will revisit the deploy path in a bit. so currently the post build hook does the job for the deploy?

apps/backend/package.json
lorenjohnson: yep, I did peripheral research after your note yesterday, and that tracks. it seems heroku doesn't install dev dev dependencies on build--which makes entire sense. we basically have to add typescript to regular dependencies--which is weird/wrong--or we look at other solutions on a few heroku threads/tickets out there. it is a common requirement / question, but also if I'm understanding what you have now with the postbuild hook, that may just be the solution.

apps/web/package.json
lorenjohnson: it would be nice if we didn't need to do that, but as of last I checked it was still an issue. hopefully with the babel upgrades and such it isn't an issue anymore. to test make a change to an existing graphql query (probably one wrapped in `gql` or an actual `.graphql` file if we still have any ) and if the changes come through either on hot reload or restart without this extra bit, then we're golden... you probably already knew/remembered this, but just to make sure we're talking about the same thing :)

package.json
tibetsprague: ideally the postbuild uses a script that detects from an environment variable whether we are doing a frontend or back-end deploy and only builds the web for frontend deploys. but builds shared for both. that is a later task

PR #78, opened 2024-11-17
Enable @hylo/shared to work in monorepo for backend

lorenjohnson:
This is the minimal setup required to make the @hylo/shared code work with the backend. 

The good news is the module dependency and importing (actually "require"ing after babel'ified) in the backend looks and work just as it does for the frontends. 

The only bad news is that the CommonJS build step is in this configuration only ran when the backend dev server is started (i.e. `yarn workspace backend dev` / `yarn backend:dev` / `cd apps/backend && yarn dev`) and on `start` as well (which captures the production case). There is no "watch". You need to restart the dev server to see changes reflected in @hylo/shared code in backend dev. To add a concurrent build/watch step isn't a big additional, but for there were a couple issues there which I would rather leave to look at later if it proves to be a hinderance. 

_From here--in other PRs--I'll be adding another shared package or two for URQL / graphql related stuff, but these libraries I continue to think that these libraries won't be of interest to the backend, so will just use the same configuration as @hylo/shared in the frontends which will give us watch and raw ESM usage..._


thomasgwatson (approved):
lgtm

tibetsprague (approved):
this does look good but im nervous about deployments again, because the only way I got the front-end to work is by deploying it from the root directory and having that just do a front-end workspace deployment. with this there will be a whole other round of trying to figure out how to deploy both the front-end and the back-end when they are both using workspaces 😬

lorenjohnson:
> this does look good but im nervous about deployments again, because the only way I got the front-end to work is by deploying it from the root directory and having that just do a front-end workspace deployment. with this there will be a whole other round of trying to figure out how to deploy both the front-end and the back-end when they are both using workspaces 😬

I hear ya. I would like to take a pass on CI and deployments after I get just a bit further on Mobile URQL. In this case the alternatives would all look like basically versioning, building, and publishing the @hylo/shared package, as we did before, or otherwise building and storing the CommonJS transpile of the library in a version which the Backend relies upon. We'll figure-out any deployment issues, and I'm hopeful it will get as smooth / smoother than ever by the time some other dust settles, but having to remember to keep a static build of @hylo/shared stored in the repo up-to-date comes with more risks. 

I'll merge for now, and once we get to the polishing rounds on CI and Deployment I can re-evaluate any other ways.

PR #97, opened 2024-12-05
Fix moderation action notifications

tibetsprague:
Emails and DMs


apps/backend/api/models/ModerationAction.js
```diff
@@ -121,22 +82,71 @@ module.exports = bookshelf.Model.extend({
     return action
   },
 
+  async sendEmailsForModerationAction ({ reporterId, postId, groupId, type = 'created' }) {
```
thomasgwatson, line 85: Ahh so I had this as an instance method instead of a class method?

thomasgwatson (approved):
lgtm

PR #101, opened 2024-12-16
Fix all web tests! Most snapshot tests gone.

tibetsprague:
Everything using react testing library instead of jest

Required adding a lot more accessible aria attributes to our components.

Some more components converted to function components

Comment editing fixed
Reactions view on user profile fixed
Fix Group welcome modal
Fix display of Big Grid Items
Fix Start Group button in Drawer
Fix related groups settings tab


thomasgwatson:
🎉🎉🎉🎉🎉🎉🎉🎉🎉

PR #106, opened 2024-12-26
Only add /post/ when its not already coming from a post view url

### LINKED ISSUE #105: Post Detail not loading: some instances of incorrect URL being generated
Seems like there are some instances where the URL is malformed for the post detail view
https://www.hylo.com/all/post/post/81530

thomasgwatson:
closes #105


tibetsprague (commented):
I dont think this quite the right place to fix this. baseUrl should not include the original /post/id, so it broke somewhere else in here

PR #108, opened 2024-12-27
Support for publishing Murmurations profiles for public groups 

tibetsprague:
Add a new toggle to Public groups to turn on publishing a group profile to the https://murmurations.network index. Every time the group is updated it re-publishes the public info to the inbox. The murmurations profile can be found at https://hylo.com/noo/group/group-slug/murmurations. 

This will be used to enable interoperable ecosystem map creation where organizations can use their Hylo profile as their main online presence that automatically updates as the group changes. 

We are being paid for this work by the Collaborative Tech Alliance.


apps/backend/api/controllers/MurmurationsController.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  // post: async function (req, res) {
```
tibetsprague, line 2: this part is coming soon

thomasgwatson (approved):
lgtm

lorenjohnson:
@thomasgwatson and @tibetsprague this now has conflicts to resolve, but otherwise can/should this be merged? I am not familiar with the feature.

tibetsprague:
I actually just learned that there is one more piece that needs to be added to this before we want to merge it in

PR #110, opened 2024-12-30
My context menu

thomasgwatson:
Screenshot 2024-12-30 at 5 29 17 PM


apps/web/src/util/navigation.js
```diff
@@ -146,8 +146,8 @@ export function messagePersonUrl (person) {
 }
 
 // Person URLs
-export function currentUserSettingsUrl (view = '') {
-  return '/settings' + (view ? '/' + view : '')
+export function currentUserSettingsUrl () {
+  return '/my/edit-profile'
```
tibetsprague, line 150: i think this URL should be /my/settings and that you should still be able to pass in a view to go to the specific settings page. I know we need this for resetting password for example. though we will need to update that URL to point to /my/settings anyway

apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -420,11 +420,16 @@ export default function AuthLayoutRouter (props) {
               <Route path='groups/:groupSlug/*' element={homeRoute} />
               <Route path='post/:postId/*' element={<PostDetail />} />
               {/* **** My Routes **** */}
-              <Route path='my/:view/*' element={<Stream context='my' />} />
+              <Route path='my/posts/*' element={<Stream context='my' />} />
```
tibetsprague, line 423: we may run into an issue losing the :view parameter from the URL params? not sure. and there may be other ways to handle that

apps/web/public/locales/es.json
```diff
@@ -1101,7 +1101,7 @@
   "this role/badge?": "este rol/insignia?",
   "title: text": "título : texto",
   "to set it up.": "para configurarlo.",
-  "topics: up to 3 topic names separated by spaces and/or commas e.g. “food organic”": "temas: hasta 3 nombres de temas separados por espacios y/o comas, p. \n“comida organica”",
+  "topics: up to 3 topic names separated by spaces and/or commas e.g. “food organic”": "temas: hasta 3 nombres de temas separados por espacios y/o comas, p. \n"comida organica"",
```
tibetsprague, line 1104: this looks broken

tibetsprague (changes requested):

apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.js
thomasgwatson: I passed them in as props instead, in this instance

apps/web/src/util/navigation.js
thomasgwatson: One clarification:
Previously it was urls like 
/settings/notifications
/settings/invitations

And now it is 
/my/notifications
/my/invitations

As for changes to the function:
I just simplified it because it was never used with a param but could just switch it to include adding /notifications, etc

As for other parts of the platform that rely on: `/settings/*`:
These paths still route to the UserSetting component like before. I didn't want to force an update to mobile right away with this change, for example. So other settings urls should still work the same

apps/web/public/locales/es.json
thomasgwatson: yeah, cursor was doing this weird thing when I was asking it to translate completely unrelated lines; fixed it again

apps/web/src/util/navigation.js
tibetsprague: gotcha. Hmm. well then i think /my/edit-profile should by /my/settings. I'm also on the fence about whether it should be /my/notifications or /my/settings/notifications. I'm thinking about group settings too - is it /group/slug/privacy or /group/slug/settings/privacy. I think the latter is more clear and also because of how group settings pages will not all live in the context menu but will be an additional menu layer deep it will probably all need to be underneath that /settings path. and then i wonder if user settings should be similar. They don't need to be since they will always be visible in the context menu, but it would be nice to have them conceptually feel similar. The settings all live under .../settings/... which is the view, instead of each settings page is now its own view 🤔 What do you think?

we will also need redirects for at least some of the old URLs, specifically the /settings/account one, to /my/account

thomasgwatson: Currently, in lieu of redirects, those routes are still valid and point at the same destination. Or are you suggesting we eliminate any use of the old urls?

For the userSettings, I've stripped out the extra menu; its redundant, since we have each menu option covered in the my context. That doesn't hold true for group settings, at least for now; I don't think its worth it to take up all that group menu real-estate (even if only admins see it).

As I see it, we are sitting with the question of whether we make the user settings urls conform more to the generic structure: `/:context/<optionalId>/:view` or make them conform to the sub-norm of the '/settings/' urls, by throwing /settings/ in there, and adding another level of the hierarchy (so maybe `/:context/<optionalId>/:view/:subview`. The first option allows us to treat the menu items for settings and their views like the overall new url pattern. The second requires a little more futzing and a few more conditionals here and there. I'm inclined to favor using the more generic url pattern, rather then create an exception case with a class of two (instead of one exception, with group settings)

If the desire is to make the url more human readable, i'd opt for `/my/notifications` to become `/my/notification-settings` vs `/my/settings/notifications`, again, to conform to the wider url norm

tibetsprague: We should definitely set up the redirects and get rid of the old URLs and components as much as we can, commit to the new way. After further thought I do feel that /my/settings/notifications is the cleanest, most standard way of doing the URLs, especially if all the settings stuff is going to continue sharing code and live in a shared settings directory. Having nested routes like that does in theory work really well with react router, you just add the second Route path='notifications' under a Route that has path='settings/*', or we just use one Route path='settings/notifications', I'm not sure what conditional would be required. But I'm also ok with just /my/notifications if you really feel thats better (I'm not into not /my/notification-settings though).

PR #115, opened 2025-01-03
Mobile+web merge

thomasgwatson:
This branch includes URQL-POC and web/new-design

I got web, ios and android off this branch.

There were limited merge conflicts, mainly just futzing around with the yarn.lock. Needed to run yarn dedupe to clean up something.

Also needed to add <BrowserRouter> as a provider. Tibet, I saw this was added in the testing setup but not to the regular app hierarchy. I'm surprised it complained about this after the merge but not before; I guess there was a little dependency bump that shifted things?

Not really suggesting me merge this to dev but just created a PR to leave this note here


lorenjohnson:
As the urql-poc branch is already up-to-date with dev I went ahead and re-targeted this PR to urql-poc so the diff is the review of what is changing on that branch, vs the entire urql conversion + new design. I hope that makes sense. It certainly helps me to review, but also I think is the right call as I think we the URQL stuff should be reviewed in a PR another time and on its own. 

Also, I think merging the new design work into the urql-poc branch now is the right action, as I don’t expect the URQL converted app to be released until the new design work is also ready to release?  That could be wrong, but seems true to me right now. Once merged then we can review the urql-poc branch against dev and go ahead and merge it down there and we all continue our work with that as the base branch from that point.

Merging the new design work may break some Mobile integration of Web, but it shouldn’t slow me down much right now as those spots aren’t much in focus in the URQL conversion. Either way it will be good for us to be aware of any issues in the integration sooner than later. Maybe this is too soon, but it seems like this may more or less be a good time to start working from a shared base branch.

thomasgwatson:
Yeah, I figure we throw it all on the same branch. We won't be merging down to dev anytime soon.

Tried to be somewhat well-behaved, regarding breaking of web views for mobile. For example, I've moved the user settings UI to point at a different url but the old urls for user settings should still work. We should probably maintain those settings urls until well after the redesign release, since folks that don't update their mobile apps will still depend on the old settings urls.

thomasgwatson:
@tibetsprague assuming this PR/branch is now redundant and closing

PR #116, opened 2025-01-04
fix(web): HyloEditorMobile moves to create ready instead of onBeforeCreate, fixes #109

### LINKED ISSUE #109: Commenting on a sub-comment breaks the comment box
Clicking on the reply to a comment that is already a child/sub comment of someone else's comment brings up the comment box without the ability to type into it

lorenjohnson:
fix #109

### DIFF
--- a/apps/web/src/components/HyloEditor/HyloEditor.js
+++ b/apps/web/src/components/HyloEditor/HyloEditor.js
@@ -22,7 +22,7 @@ const HyloEditor = React.forwardRef(({
   onAddLink,
   onAddMention,
   onAddTopic,
-  onBeforeCreate = () => {},
+  onCreate = () => {},
   onUpdate,
   onEnter,
   onEscape,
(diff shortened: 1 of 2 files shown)


apps/web/src/components/HyloEditor/HyloEditor.js
```diff
@@ -34,100 +34,124 @@ const HyloEditor = React.forwardRef(({
...
-          return {
-            Enter: ({ editor }) => {
-              if (!onEnter) return false
-              return onEnter(editor.getHTML())
-            },
-            Escape: () => {
-              if (!onEscape) return false
-              return onEscape()
-            }
+  const extensions = [
```
lorenjohnson, line 38: Moved extensions into their own variable for clarity, but not changes in this area

apps/web/src/components/HyloEditor/HyloEditor.js
```diff
@@ -34,100 +34,124 @@ const HyloEditor = React.forwardRef(({
...
+    onCreate,
     onUpdate: ({ editor }) => {
       if (!onUpdate) return
-
       onUpdate(editor.getHTML())
     }
-  }, [placeholder, contentHTML]) // TODO: changing the placeholder resets the content of the editor which is probably not what we want
+  })
+
+  // Dynamic setting of initial editor content
```
lorenjohnson, line 131: Changes how we set initial content, no longer re-creating the editor instance. This was a necessary part of this change, but is a better practice at this stage with TipTap.

apps/web/src/components/HyloEditor/HyloEditor.js
```diff
@@ -34,100 +34,124 @@ const HyloEditor = React.forwardRef(({
...
+  useEffect(() => {
+    if (editor.isInitialized) {
+      editor.commands.setContent(contentHTML)
+    }
+  }, [editor?.isInitialized, contentHTML])
+
+  // Dynamic placeholder text
```
lorenjohnson, line 138: As above, but for Placeholder text.

apps/web/src/components/HyloEditor/HyloEditorMobile.js
```diff
@@ -101,7 +100,7 @@ export default function HyloEditorMobile (props) {
     className: 'hyloEditorMobile',
     onUpdate: handleUpdate,
     onEnter: handleEnter,
-    onBeforeCreate: handleBeforeCreate,
+    onCreate: handleCreate,
```
lorenjohnson, line 103: This is the key change, and only impacts HyloEditorMobile. Previously, likely because of different behaviour in an earlier version of TipTap, we handled setting the "loaded" state, which is the only thing we use this event for, via `onBeforeCreate`. That was too early now, so now hooking into `onCreate` to set editor loaded which HyloEditorMobile pays attention to to know when it can send editor updates through the WebView.

tibetsprague (approved):

thomasgwatson (approved):
lgtm

PR #132, opened 2025-01-13
Better datetime picker

### LINKED ISSUE #70: Better date/time picker
Make it easier to type in the time when creating a post (dont jump around, understand the time better)

KevinTriplett:
closes #70 

Added ShadCN ui expansion https://shadcnui-expansions.typeart.cc/docs/datetime-picker
Added tailwind css z-index 'z-200' in order to show the calendar above the modal window
Note: z-200 added to ui select, which might affect other usage
Added Luxon datetime library
Event endTime tracks changes to startTime, to reduce frequency of error message

Feature suggestion: add warning when selecting an event date in the past


KevinTriplett:
Note: I'll create a separate PR that addresses the annoying error when creating an event and choosing a start date/time

KevinTriplett:
Note: I'll create a separate PR that addresses the annoying error when creating an event and choosing a start date/time

KevinTriplett:
Ready for review

Note: initializing end date/time based on start date/time is problematic and needs help

KevinTriplett:
This last bug was unexpected and I may report it to the component maintainers. When changing the calendar using the left/right chevrons (< >) it fired the onChange handler which simulated the user picking a date from the calendar. Yikes!

apps/web/package.json
```diff
@@ -139,6 +140,7 @@
     "lodash": "~4.17.21",
     "lru-cache": "^4.1.1",
     "lucide-react": "^0.453.0",
+    "luxon": "^3.5.0",
```
tibetsprague, line 143: im a little hesitant to add/use Luxon until we are fully replacing moment with it, otherwise we have both filling up our bundle size. is it hard to use moment here for now?

apps/web/src/components/PostEditor/PostEditor.js
```diff
@@ -243,6 +245,16 @@ function PostEditor ({
     }
   }, [])
 
+  const calcEndTime = (startTime) => {
+    var msDiff = 3600000 // ms in one hour
```
tibetsprague, line 249: use let instead of var. your linter should have caught that one. you should see lint errors in VS Code and in the terminal where the front-end server is running

apps/web/src/components/ui/select.jsx
```diff
@@ -55,7 +55,7 @@ const SelectContent = React.forwardRef(({ className, children, position = 'poppe
     <SelectPrimitive.Content
       ref={ref}
       className={cn(
-        'relative z-50 max-h-96 min-w-[8rem] overflow-hidden rounded-md border bg-popover text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
+        'relative z-200 max-h-96 min-w-[8rem] overflow-hidden rounded-md border bg-popover text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
```
tibetsprague, line 58: 200 seems like a lot, and we certainly dont want every select to have that. why not just make it 60? and also just pass it in to the specific select?

apps/web/src/components/ui/calendar.jsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
+import { ChevronLeft, ChevronRight } from 'lucide-react'
+import { DayPicker } from 'react-day-picker'
+import { cn } from '@/lib/utils'
+import { buttonVariants } from '@/components/ui/button'
+
+function Calendar ({ className, classNames, showOutsideDays = true, ...props }) {
```
tibetsprague, line 7: are we using this calendar component yet?

tibetsprague (changes requested):

apps/web/src/components/ui/select.jsx
KevinTriplett: 200 is the lowest value that made the calendar visible. I'll look at decreasing the model z-index.

apps/web/src/components/ui/calendar.jsx
KevinTriplett: Good catch -- I added it to the project because it said to. I'll remove it.

apps/web/src/components/PostEditor/PostEditor.js
KevinTriplett: Will change it, thanks!

apps/web/src/components/ui/select.jsx
KevinTriplett: Huh ... the modal is at z-index: 10 so that's not it. There must be something I'm not seeing.

apps/web/src/components/PostEditor/PostEditor.js
```diff
@@ -246,13 +246,11 @@ function PostEditor ({
...
-    var msDiff = 3600000 // ms in one hour
+    let msDiff = 3600000 // ms in one hour
     if (currentPost.startTime && currentPost.endTime) {
-      const start = DateTime.fromJSDate(currentPost.startTime)
-      const end = DateTime.fromJSDate(currentPost.endTime)
-      msDiff = end.diff(start)
+      msDiff = moment(currentPost.endTime) - moment(currentPost.startTime)
     }
-    return DateTime.fromJSDate(startTime).plus({millisecond: msDiff}).toJSDate()
+    return new Date(moment(startTime) + msDiff)
```
tibetsprague, line 253: you can just create a new Date from a moment object? thats wild

KevinTriplett: Yeah, wild, because moment + number = milliseconds, which js Data happily takes for initialization.

I don't know if js Date accepts a moment object directly.

apps/web/src/components/ui/select.jsx
KevinTriplett: Resolved with below z-index commit

tibetsprague (approved):

PR #144, opened 2025-01-22
Adds Backend subscriptions for WebSockets events, and applies them to Mobile

lorenjohnson:
# Features summary

-   Adds three subscriptions:
    -   `updates`:  Streams new Messages, MessageThreads, and Notifications for the current user through a new Update grapqhl union type.
    -   `peopleTyping`: Along with the paired peopleTyping mutation and `usePeopleTyping` hook which should be re-usable in Web, handles typing state. Is applied to either a messageThreadId, postId, or commentId.
    -   `comments`: Streams new comments for a given postId or commentId (implicitly a parentCommentId). Currently applied only in PostDetails on Mobile for parent comments (no UI impact at the moment), perhaps has other possible near-term applications.
-   Uses HTTP/2 Server Side Events (SSE) vs WebSockets, which seems to be the preference of the community these days. SSE support on React Native is achieved through a poly (pony?) fill and seems to work, but is something I'm watching and giving this a bit more attention is part of follow-up.
-   There remains badge updating on Mobile which should be pretty simple :crossed_fingers::skin-tone-3:(https://a.slack-edge.com/production-standard-emoji-assets/14.0/apple-medium/1f91e-1f3fc@2x.png), and refinement to the underlying along the way as we polish things this month, but working well as is.

## Focus recommendations for review

For the review there is a lot to pay attention to, but recommended focus is:

- [ ] The subscription setup in the backend. Mostly `api/graphql/index.js` and `api/graphql/makeSubscriptions.js`.
- [ ] The application of `useSubscription` in mobile in `AuthRootNavigator`, `Thread`, `usePeopleTyping`, and `PostDetails`.
- [ ] The current state of `apps/mobile/urlq-shared/updates.js` with regard to the Subscriptions.
- [ ] The bridging between WebSocket isTyping and the `peopleTyping` subscription and mutation (small fix on Web for Messages isTyping)

---

**Additional detail and follow-on development notes**

Resolving Mobile UI badge updates and any fine-grained URQL caching updates are out-of-scope for this branch.

Messages conversion status:

- [x] commentAdded
- [x] messageAdded ("updates" subscription)
- [x] userTyping ("peopleTyping" mutation + subscription)
- [x] newThread ( ("updates" subscription)
- [x] newNotification  ("updates" subscription)
- newPost (or group subscription) -- will save for later consideration / iterations
- Badge updating and possibly some minor URQL cache reconciliation remains for incoming updates, and will be left out of scope for this branch

Remaining dev todos:

- [x] Relay sockets event for peopleTyping from backend subscription so peopleTyping appears on Web
- [x] Remove unused subscriptions
- Follow-up PR a bit later: 
  - Look into a Graphql Yoga upgrade to get us on Single Connection vs Distinct Connect Server Side Events (SSE) for subscriptions. After initial research  upgrade from v2.13 that we're on all the way to the current v5.x looks like it could go very quickly.
  - Test and refine reconnect/retry for SSE


lorenjohnson:
> cool!

@tibetsprague aH yes, I kinda accidentally marked for review early. Will assign you and Tom again once it is ready ready sometime in the next day or so. 

There is a bit more refinement coming with the final message implementations, but it's mostly here. It was pretty rough getting through the layers, but once things started working it became pretty satisfying. May still switch to WebSocket underneath instead of HTTP2/Events, not sure yet. It should mostly just be an adapter change if I do though. There are definitely some wins here, which like a lot of this stuff, we'll feel more later.

apps/backend/api/graphql/makeSubscriptions.js
```diff
@@ -0,0 +1,79 @@
...
+// ]
+
+/**
+ * Filters out subscription events where the current user is the creator.
+ * @param {Object} options - Options for filtering.
+ * @param {Object} options.context - The GraphQL context containing `currentUserId`.
+ * @param {Function} [options.getter] - Optional custom getter to extract the creator ID.
+ * @returns {Function} - A function that filters subscription payloads.
+ */
+const withDontSendToCreator = ({ context, getter } = {}) => {
```
thomasgwatson, line 21: Mmm, does this create duplicate records or something? Or is the assumption that the originator will invalidate their cache and cause a refresh anyway?

lorenjohnson: No in the case of an URQL client with graph cache it dedupes by default, just best practice to not send back events to the client that don't belong. The mutation return is how the creator gets updated.

lorenjohnson: Please do keep perusing and such, so very welcome of course, but just to be clear this PR is not ready for any sort of review quite yet. I'll move it out of draft and assign again once it is. Also quite a lot more still changing / to be pushed on this one.

apps/backend/docs/20250124 - Proposal - AutoSubscriptionsForMutations.md
lorenjohnson, line 1: This is just a genAI digestion of some code I started with and didn't want to throw out as there was some good learning in it. As a proposal I'm not particularly interested in it actually, but as an illustration of how to use graphql Envelop middleware with our stack it I think is helpful to keep a trace on what I found. That said, I'll probably remove this before I'm done as I'd rather having some notes in this directory about remaining (eventual) Backend projects that I do care about: Moving to full Module/import vs require mode, and some substantial clarifying/simplifying and improving our graphql server setup.

apps/mobile/src/urql-shared/client.js
```diff
@@ -20,10 +22,60 @@ const cache = cacheExchange({
 })
 
 const client = createClient({
+  // fetchSubscriptions: true,
```
thomasgwatson, line 25: Keeping this commented out?

lorenjohnson: Ah thanks, yes I think I'll remove it. Worth noting that this will probably be on and all we do to handle the Web URQL client subscriptions connection on Web, but didn't support React Native.

apps/mobile/src/navigation/TabsNavigator.js
```diff
@@ -56,7 +56,7 @@ export default function TabsNavigator () {
     <Tabs.Navigator {...navigatorProps}>
       <Tabs.Screen name='Home Tab' component={HomeNavigator} />
       <Tabs.Screen name='Search Tab' component={SearchNavigator} />
-      <Tabs.Screen name='Messages Tab' component={MessagesNavigator} />
+      <Tabs.Screen name='Messages Tab' component={MessagesNavigator} options={{ tabBarBadge: 999 }} />
```
thomasgwatson, line 59: Whats this do?

apps/mobile/src/urql-shared/updates/index.js
```diff
@@ -71,5 +71,32 @@ export default {
...
+  Subscription: {
+    comments: (result, args, cache, info) => {
+      console.log('!!!!! comments - result, args, info:', result, args, info)
+    },
+    updates: (result, args, cache, info) => {
+      const update = result?.updates
+
+      switch (update?.__typename) {
+        case 'Message': {
+          console.log('!!!! updates TODO: new Message. Increment Messages tab badge', result, args)
```
thomasgwatson, line 84: ??

thomasgwatson (approved):
lgtm

apps/mobile/src/urql-shared/updates/index.js
lorenjohnson: As mentioned in the perhaps tl;dr description:), managing badges is out of scope in this PR and "coming soon". The logging, and the 999 badge on the Messages tab are meant as book marks.

apps/mobile/src/navigation/TabsNavigator.js
lorenjohnson: A bookmark for badges to come. Quite unmissable in the UI

tibetsprague (approved):
get it

PR #148, opened 2025-01-23
Adds StandardJS linter to global package, in backend some back linting, and moves from requires to imports for yoga-graphql

lorenjohnson:
This is all superficial, I promise nothing to fret about. 

- The only things that could have a consequence is the StandardJS move to global, which should be the right direction for our monorepo but may have an edge or two to file off. For now works great for making VS Code linting happy, and does ensure we're on the same version of Standard across all code bases. 

- Switches from requires to imports for `@graphql-yoga/node` in the Backend. Because we can (earlier versions were presumably not Modules compatible), and eventually if we've moved all our requires to imports we can switch the Backend to modules which would simplify our shared code, amongst other things. 

- Some auto lint fixes applied to some Backend code on files I was already touching.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@hylo/shared": "workspace:*"
   },
   "devDependencies": {
+    "standard": "^17.1.2",
     "typescript": "5.0.4"
   },
   "resolutions": {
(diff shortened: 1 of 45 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
     "@hylo/shared": "workspace:*"
   },
   "devDependencies": {
+    "standard": "^17.1.2",
```
lorenjohnson, line 31: This, and the removals of standard from the individual packages is what to pay attention to here. Everything else has no functional impact.

tibetsprague (approved):

PR #155, opened 2025-01-24
Post body proposal to urql mobile

thomasgwatson:


apps/mobile/src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -96,19 +100,19 @@ export default function PostBodyProposal ({
   function handleVote (optionId) {
     if (votingMethod === VOTING_METHOD_SINGLE) {
       if (currentUserVotesOptionIds.includes(optionId)) {
-        dispatch(removeProposalVote({ optionId, postId: id }))
+        removeProposalVote({ optionId, postId: id })
       } else if (currentUserVotesOptionIds.length === 0) {
-        dispatch(addProposalVote({ optionId, postId: id }))
+        addProposalVote({ optionId, postId: id })
```
lorenjohnson, line 105: You can just `const { fetching: addProposalVoteMutationFetching }` or above in `const [{ fetching: addProposalVoteMutationFetching, addProposalVote] = useMutation(addProposalVoteMutation)`, as you wish. I would recommend a `const loading = useMemo(() => addProposalVoteMutationFetching && addProposalVoteMutationFetching && addProposalVoteMutationFetching, [addProposalVoteMutationFetching, addProposalVoteMutationFetching, addProposalVoteMutationFetching]` or similar and then using that loading state in the component to block actions until they're resolved. Even with an updater or optimistic updater that reconciles to the cache without and invalidate, it can still take a visual period of time so best to handle the UI blocking part directly in the component here.

apps/mobile/src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -135,7 +139,7 @@ export default function PostBodyProposal ({
...
+          <TouchableOpacity key={`${option.id}+${currentUserVotesOptionIds.includes(option.id)}`} style={[styles.proposalOption, votingComplete && styles.completed, proposalStatus === PROPOSAL_STATUS_DISCUSSION && styles.discussion, proposalStatus === PROPOSAL_STATUS_VOTING && styles.voting, proposalStatus === PROPOSAL_STATUS_CASUAL && styles.casual, currentUserVotesOptionIds.includes(option.id) && styles.selected, votingComplete && styles.completed, votingComplete && highestVotedOptions.includes(option.id) && styles.highestVote]} onPress={isVotingOpen(proposalStatus) && !isVoting && !votingComplete ? () => handleVote(option.id) : () => {}}>
```
lorenjohnson, line 142: Please put this logic in a const (or a `useMemo` call if needed) above. I apologize ahead of time if this is still in progress, and can wait to review until it's marked ready if so. Just please do mark it to ready when it is if that is the case. 

_One more comment while I'm here though, and this was my fault for misdirecting you:

`addProposalVoteMutationFetching` should be `addProposalVoteFetching`, and the same for the other two. We drop the `*Mutation*` part once we're outside of graphql land and in component land if that makes sense._

apps/mobile/src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -77,8 +77,12 @@ export default function PostBodyProposal ({
   id
 }) {
   const { t } = useTranslation()
-  const dispatch = useDispatch()
 
+  const [{ fetching: addProposalVoteMutationFetching }, addProposalVote] = useMutation(addProposalVoteMutation)
+  const [{ fetching: removeProposalVoteMutationFetching }, removeProposalVote] = useMutation(removeProposalVoteMutation)
+  const [{ fetching: swapProposalVoteMutationFetching }, swapProposalVote] = useMutation(swapProposalVoteMutation)
+
+  const isVoting = useMemo(() => addProposalVoteMutationFetching && addProposalVoteMutationFetching && addProposalVoteMutationFetching, [addProposalVoteMutationFetching, addProposalVoteMutationFetching, addProposalVoteMutationFetching])
```
lorenjohnson, line 85: Sorry, my bad in my too quick notes earlier. This should simply be:

const isVoting = addProposalVoteFetching && removeProposalVoteFetching && swapProposalVoteFetching

Note the renames and reflect them starting with the `fetching: addProposalVoteMutationFetching` part, and no need for `useMemo` here.

lorenjohnson (approved):

PR #156, opened 2025-01-24
New notifications experience

tibetsprague:
Per post notifications and email digests are now controlled on a per group basis rather then globally

New notification settings UI


apps/backend/api/models/Group.js
```diff
@@ -1043,7 +1045,7 @@ module.exports = bookshelf.Model.extend(merge({
   async doesMenuUpdate ({ groupIds, post, customView, groupRelation = false }) {
     if (!post && !customView && !groupRelation) return
     console.log('broken here post', post)
```
lorenjohnson, line 1047: remove console.log (unless this is still draft then nvm :) )

apps/web/src/components/ui/info.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react'
```
lorenjohnson, line 1: Just curious in catching-up to Web a bit while walking through this change set, why ui and why *.jsx in this area?

apps/web/src/routes/UserSettings/NotificationSettingsTab/NotificationSettingsTab.js
```diff
@@ -1,255 +1,212 @@
...
-    const { messageSettings, updateUserSettings } = this.props
-    const newMessageSettings = {
-      ...messageSettings,
+// Generic function to get current settings for any notification type
+const getCurrentSettings = (me, settingKey) => convertEnumToFlags(me.settings?.[settingKey])
+
+const getAllGroupsSettings = createSelector(
+  getMyMemberships,
+  memberships => ({
+    sendEmail: every(m => m.settings && m.settings.sendEmail, memberships),
```
lorenjohnson, line 57: Not important, but curious if there was a reason you didn't use optional chaining (e.g. `m?.settings?.sendPushNotifications`) here and in the line below?

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -533,53 +533,53 @@ function ListItemRenderer ({ item, rootPath, groupSlug, canDnd, isOverlay = fals
...
-            );
-          } else if (rootPath === "/my" || rootPath === "/all" || rootPath !== "/members" || (item.title && item.type !== "chat")) {
+            )
+          } else if (rootPath === '/my' || rootPath === '/all' || rootPath !== '/members' || (item.title && item.type !== 'chat')) {
             return (
               <MenuLink
                 to={itemUrl}
-                externalLink={item?.customView?.type === "externalLink" ? item.customView.externalLink : null}
-                className="text-base text-foreground border-2 border-foreground/20 hover:border-foreground/100 hover:text-foreground rounded-md p-2 bg-background text-foreground mb-[.5rem] w-full transition-all scale-100 hover:scale-105 opacity-85 hover:opacity-100 flex align-items justify-between"
+                externalLink={item?.customView?.type === 'externalLink' ? item.customView.externalLink : null}
```
thomasgwatson, line 573: In general, I plan to go over all of this and double-check if things can be refractored from aarons changes

thomasgwatson (approved):
lgtm

apps/web/src/components/ui/info.jsx
thomasgwatson: We are switching all components over to jsx, although I can't recall why Tibet has gone with this.

/ui is just the place where shadcn npx commands drops those components. Just helping us keep track of what's been pulled in from shadcn already. At some point we'll go through and weed out defunct components and decide if we want to hold on to the /ui

apps/backend/api/graphql/schema.graphql
lorenjohnson, line 1: These changes here (mirrored in the MeQuery changes in Web) need to be reflected in Me queries in Mobile as well so this doesn't break Mobile on merge. However, I don't think there is any UI which needs changing at this time.

lorenjohnson (changes requested):
Needs Mobile `me` query update as well, as per note on `schemes.graphql`. Rest of comments are trivial or just curiosities.

apps/backend/api/models/Group.js
tibetsprague: there is actually still a bit more work to do here, but whats here is reviewable. thanks

apps/web/src/components/ui/info.jsx
tibetsprague: because then the editor knows its jsx and not just js. its needed for certain linting rules e.g.

apps/web/src/routes/UserSettings/NotificationSettingsTab/NotificationSettingsTab.js
tibetsprague: just copying old code

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
tibetsprague: yeah, I was going to go over better use of MenuLink with him at some point too

apps/backend/api/graphql/schema.graphql
tibetsprague: hmh, why are those notification settings even being loaded on mobile? I dont think they are being used anywhere since the settings pages are webviews

lorenjohnson: they aren't, but the queries have been in sync more or less for some time (and I imagine continuing in that direction more not less). either way, when this gets removed from the schema and our query on mobile still queries these fields it will crash.

apps/backend/api/models/Group.js
lorenjohnson: Gotcha. One workflow that I like is to keep things in draft, and if we're doing a in process review then we still can request review and the draft status conveys this. when its not in draft I'm reviewing as if its going to production. come along with me on not that if its helpful, if not no thing and I'll adapt :)  

It's up for me now as I am thinking a PR process will be pretty helpful to keeping some clarity in the coming weeks.

apps/web/src/components/ui/info.jsx
lorenjohnson: Needless to day this is a long standing debate in the community of course, with the React team somewhat being officially not fans of JSX while not saying it loudly. I probably have a light preference for plain .js, but certainly not a strong one. I imagine in Vite it's helpful for quicker hot reloading. Good to know either way. Are you wanting us to do the same with mobile files? You can answer later, but the question is now in me :)

apps/backend/api/graphql/schema.graphql
tibetsprague: to me it makes more sense to only query for what we actually want or need in each case. you think we should keep the queries in sync because maybe in the future we will want? or because we will share query code?

apps/backend/api/models/Group.js
tibetsprague: yeah I agree, should keep things as drafts. In this case I felt like the work done was reviewable and so I didnt keep it as a draft. the other work will just add on top

apps/web/src/components/ui/info.jsx
tibetsprague: interesting, I didnt know there was controversy. yes exactly for vote hot reloading each component needs to be in its own file with only 1 default export. and I have a listing rule for that but I want it to only loop at component files so thats why having them as .jsx makes things much easier.

lorenjohnson: Yes, I remembered you'd saiid that and imagined this came from the same root... As for the controversy, it's really not that important, but since you kinda asked :). Enjoy the read 🫠 : https://github.com/airbnb/javascript/pull/985#issuecomment-239145468

thomasgwatson: for fwiw, Dan now is working primarily on the React Native bluesky app, and all their files are .tsx

PR #159, opened 2025-01-25
chore(mobile): Fixes various issues with changing groups, adds Zustand dependency

lorenjohnson:
This is a weird set of things that all came-up during debugging navigating between Context Group (My, Public, All) and normal groups. There were some persistent React errors and crashes coming up around hook state (returning before all hooks ran). This fixes those. 

Also this branch adds Zustand as a dependency which I now use as a local store for currentGroupSlug (useCurrentGroup, useCurrentGroupSlug, etc). It is much easier to work with than our Redux setup, even in the bare non ReduxORM state, and I intend to convert the other bits of state still in Redux to Zustand useState bit by bit. I do think it will be useful for some work coming up. There is no issue with having it in place while keeping Redux as it is for now.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -109,7 +109,8 @@
     "socket.io-client": "2.3.1",
     "tailwindcss": "^3.4.17",
     "tinycolor2": "^1.6.0",
-    "urql": "^4.2.1"
+    "urql": "^4.2.1",
+    "zustand": "^5.0.3"
   },
   "devDependencies": {
     "@babel/core": "^7.25.2",
--- a/apps/mobile/src/components/GroupWelcomeCheck/GroupWelcomeCheck.js
+++ b/apps/mobile/src/components/GroupWelcomeCheck/GroupWelcomeCheck.js
@@ -1,37 +1,29 @@
 import { useEffect } from 'react'
-import { useQuery } from 'urql'
 import { useNavigation } from '@react-navigation/native'
 import useCurrentUser from 'hooks/useCurrentUser'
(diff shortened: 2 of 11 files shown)


lorenjohnson:
FYI I kinda need the fixes on useCurrentGroup in the subscriptions work, so will be merging this up into my subscriptions branch, but happy for any review and can merge up anything that may change.

lorenjohnson:
@thomasgwatson this branch is a better branch point than dev if you're jumping back into things at the moment especially if you dig in on the Welcome Home stuff. Awaiting review/s but expecting this work to go forward.

tibetsprague (approved):
I was thinking about using zustand recently, but didnt want to jump on to something new. trying to remember what that was for... I think it was for setting header information per view. I ended up using react context but zustand probably would have been easier and cleaner

lorenjohnson:
> I was thinking about using zustand recently, but didnt want to jump on to something new. trying to remember what that was for... I think it was for setting header information per view. I ended up using react context but zustand probably would have been easier and cleaner

Yeah, it's pretty lightweight and non-obstructive. I feel entirely comfortable with dropping it in here for Mobile, and given how little Redux state is left in Mobile will almost certainly switch over entirely there by the time this is realized. In the context of Web it is "one more thing" but feels pretty close to just a `useState` that is app or whatever context wide not just the component. Very useful and ergonomic now that we're fully in functional component hooks world.

lorenjohnson:
FYI @thomasgwatson this branch is going to dev now, so you're good to branch off dev for anything you're doing now.

thomasgwatson (commented):
lgtm

PR #162, opened 2025-01-26
chore(backend): Upgrades Graphql Yoga from v2 to v5.10 (latest)

lorenjohnson:
Upgrade all the way to latest version of GraphQL Yoga went well, fairly smoothly. The jump from 2.x to 3.x was some work, but from 3.x to 5.x was mostly deprecation of compatibility with old versions of node, or otherwise relating to features we were not yet using in terms of Breaking Features. Lots of wins here. Give the new version of GraphiQL a whirl, it's quite nice and super helpful for testing Subscriptions. Otherwise the features here are all in fact backend things to be excited about like some intriguing response caching options that are mostly just "turn it on and forget it" (none of which are turned on yet). More robust persisted query support, etc. 

- Most of the lines changed here are that GraphQLYogaError is deprecated in favor of the standard GraphQLError, and we used that a lot of places.

- There are a lot of mutation resolver generator function changes in makeSchema.js#makeMutatiom (that were in graphql/index.js) to move from using to
move to getting userId from context currentUserId instead of from the userId parameter we used to provide to makeMutatioms. Didn't quite realize that basically every mutation took userId, or I may have held back on this part to keep down the delta, but it was only an hour or two of careful refactor and things were good.

- Note: Based on https://github.com/Hylozoic/hylo/pull/144 so this is 2nd in the merge queue. 

- While now supported, this upgrade does automatically turn on SSE Single Connection mode for Native, as it turns out most of what is required for that would be additional config to pool connections on the client side (Native) which I can see how to do, but reticent to add the extra complexity and will wait and see if we need it later. I will however make another much PR soon to add retry/reconnect logic to the subscriptions of Mobile as it currently doesn't reconnect if it loses connection. Then this whole direction of things should be very near a wrap...

- Let's merge this soon as any upstream changes to graphql/index.js between now and then will need to be manually reconciled due to the file split.

---

Some things for convos another time / follow-up:

- It would be probably good to take most what is in makeSchema.js now other than the schema production and indirection itself, i.e. makeAuthenticatedQueries, makeMutations, et al, and go ahead and put those in their own files. 
- Re. context.currentUser: My gut is that if we should rely on context in this way more going forward we will end-up with a backend that is more flexible/composable, easier to grok, and more testable. The resolvers will also look increasingly like the front-end update and query resolves we write in URQL, so it will be a familiar pattern (from the function signature on down). I think we could probably fairly easily end-up with the bound fetchOne and fetchMany in context so we could do away with many/most nearly all "generator" functions for our mutations. 
- For sport we might try turning on the preconfigured response caching available in this generation of Yoga. They make a bit of a fuss about it, and it might be a win. Otherwise we should keep an eye out for any Envelop plugins which could help us with anything. It's a pretty healthy ecosystem, to have our backend dependencies up-to-date helps us have easy access to what is out there. 

Zoomed out, I think what is at stake for me in our graphql setup on the backend is that it feels like the one piece of Backend architecture which, if we keep incrementally refining it, can last. It sits between our ORM/database layer (which I think is holding us back a bit), and the frontend, serving as something like a spec for both... That I think is graphql at its best.

Sorry for somewhat rambling report, tired, and about to step back for a couple days. Will keep an eye out for a PR review here though, and keep things moving if needed.


thomasgwatson (approved):
lgtm

apps/backend/api/graphql/mutations/user.js
```diff
@@ -45,31 +45,31 @@ export const verifyEmail = (fetchOne, { req }) => async (_, { email: providedEma
 
     await user.save({ email_validated: true })
 
-    req.session.userId = user.id
+    context.req.session.userId = user.id
```
tibetsprague, line 48: if we are switching to context.currentUserId everywhere then why do we set context.req.session.userId here?

tibetsprague (approved):
one question but otherwise looks good

apps/backend/api/graphql/mutations/user.js
lorenjohnson: It's a good question. We are continuing to set the actual userId of the session on the actual Express request object. The currentUserId that is in context is a copy of that value (see the createYoga#context function in graphql/index). Setting it won't reset the session to this token auth's user, but setting the req.session.userId does (evidently). We could also reset the currentUserId on context here, but it will already be reset to this value upon the next request to the server by this user.

PR #163, opened 2025-01-27
Context widgets a la mode

thomasgwatson:
Mainly just moving helper functions into shared.


apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
```diff
@@ -2,12 +2,12 @@ import React from 'react'
 import Avatar from 'components/Avatar'
 import { Grid3x3 } from 'lucide-react'
 import Icon from 'components/Icon'
-import { widgetTypeResolver } from 'util/contextWidgets'
+import { widgetTypeInferrer } from '@hylo/shared/src/ContextWidgetPresenter'
```
lorenjohnson, line 5: I would like us, for a couple of reasons for now to keep the context for these Presenters. So in this and later cases `import { ContextWidgetPresenter } from '@hylo/shared'` and then use `ContextWidgetPresenter.widgetTypeInferrer()` in the component. I simply find that easier to grok when reading a component, but I also want to be able to easily move these references around once I complete sorting out the shared stuff soon, in which case I do think these presenters will go into a front-ends only shared package where the rest of the graphql and URQL config will also go.

lorenjohnson: This is also my general preference for things coming from @hylo/shared. To restructure the top level module and then use it as context when calling the method in the component.

packages/shared/src/ContextMenuPresenter.js
lorenjohnson, line 1: The intention of *Presenters is to transform graphql types for front end usage. In mobile/src/urql-shared/presenters we currently have a GroupPresenter which has the PUBLIC, MY, and ALL context groups also defined in it. Given where we're going (and given the presence of these static contexts) I am happy to diverge on the front ends and call this ContextPresenter, but it would include all of the below and what is already in GroupPresenter.js. If you did that now I'd do it in urql-shared as I think this one isn't used anymore but mobile yet (may have missed something). Most things in urql-shared will go into a new shared package very soon. If this is also used on web go ahead and make it ContextPresenter there and update both Mobile references to GroupPresenter to whatever new path it is at. Hope that makes sense to you, ask if not.

packages/shared/src/ContextWidgetPresenter.js
lorenjohnson, line 1: So I see the trouble here and why what I'm suggesting might need more thought. I do in fact in the current GroupPresenter.js export all sorts of other little methods which have to be destructured off the import... Hmmm... Leave it like that for now then. Sorry for the confusion on that part. My remark about the ContextMenuPresenter stand though.

apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
lorenjohnson: Sorry, I was a little off there. I can see why you did it this way... Leave it be here for now.

packages/shared/src/ContextWidgetPresenter.js
```diff
@@ -0,0 +1,111 @@
...
+    case type === 'viewChat' || type === 'chat':
+      return 'chat'
+    case type === 'customView' || type === 'customview':
+      return 'custom view'
+    default:
+      return 'container'
+  }
+}
+
+// TODO redesign: create tests for this
```
lorenjohnson, line 95: Thought: These 2 todos seem possible in conflict :) Maybe if there is a refactor the testing will not be so needed?

packages/shared/src/ContextWidgetPresenter.js
```diff
@@ -0,0 +1,111 @@
...
+
+export function findHomeView (group) {
+  if (!group?.contextWidgets) {
+    throw new Error('Group has no contextWidgets')
+  }
+  const homeWidget = group.contextWidgets.items.find(w => w.type === 'home')
+  return group.contextWidgets.items.find(w => w.parentId === homeWidget.id)
+}
+
+// export function widgetTypeResolver ({ widget }) {
```
lorenjohnson, line 55: Remove this line?

packages/shared/src/ContextWidgetPresenter.js
```diff
@@ -0,0 +1,111 @@
...
+  export function widgetTypeInferrer ({ widget }) {
+  switch (true) {
```
lorenjohnson, line 57: This could be simplified, couldn't it? I don't see a set of cases I see a chain of precedence, something like that: 

```
return (
  widget?.type ||
  widget?.view ||
  widget?.viewGroup ||
  widget?.viewPost ||
  // ...etc,
  'customView' 
)
```

I woudl find that much more clear in a way. May be missing something, but if the null / boolean state of each of these is the issue perhaps that can be remedied upstream?

apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
```diff
@@ -2,12 +2,12 @@ import React from 'react'
...
 import { ViewHelpers } from '@hylo/shared'
 
 export function WidgetIconResolver ({ widget, style, className }) {
   if (!widget) return null
-  const type = widgetTypeResolver({ widget })
+  const type = widgetTypeInferrer({ widget })
```
lorenjohnson, line 10: The idea of the presenter though is that you'd have this method on the ContextWidgetPresenter and you'd transform a widget as soon as you see it like

`const widget = ContextWidgetPresenter(providedFoundQueriedWhatverWidget)`

That is the scoping I'm expecting. Is there some reason to not do that? It does have a lot of advantages in my view which I'd happily talk through with you if you see something I don't.

lorenjohnson: As you have a collection I would do the *Presenter transform on the whole collection where you can so you don't have to call the Presenter throughout the code. e.g.: 

`const widgets = mywidgetcollection.map(widget => ContextWidgetPresenter(widget))`.

You could also just make a usePresentedWidgets hook or something. The idea with Presenters is that they happen pretty high up in the component where possible, as the directly they may be going is that they are applied in URQL before the component like what happens with Reducx ORM  (just not ready to go there yet). Keeping tight on our use of them now by always transforming the whole object, and doing it early in the component will leave us in a place to switch over to that later fairly easily.

lorenjohnson (changes requested):
Going in the right directly and lot's of notes below. If we can catch each other at the right time and you're open to me getting in there with you, I think this one will be quicker to pair our way through.

apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
thomasgwatson: I was hanging on including inferredType in the presenter because I hadn't checked all the wider impacts. Done that now.

packages/shared/src/ContextWidgetPresenter.js
thomasgwatson: That works fine for .type and .view but by .viewGroup, it would return the whole group object, instead of a string. Right?

packages/shared/src/ContextMenuPresenter.js
thomasgwatson: So, I don't think we meaningfully hold these contexts as faux-groups on the front-end anymore. Like at the moment, no banner, no avatar, etc for them. 

So that might be a bit of a funny/weird point as we transition into more and more `shared` stuff

thomasgwatson: And just for `context` (oh dear)...

The ordering function here isn't universally applied to context-widgets, so I wouldn't chuck in it a presenter anyway (well, unless we include it as an optional...)

thomasgwatson: And all this stuff is currently in use by web, so I don't want to merge it with urql-shared until urql-shared has come over to `shared` land

lorenjohnson: Nevermind URQL on both the urql-shared and orderContextWidgetsForContextMenu comments. I had missed that all of the revisions were on web not mobile.

As for context groups they were never used on web I think? But we can and should share how we do it, whether that means matching Mobile with a Web pattern or vice versa. The first step for me in that is to do a potentially messy consolidation of the handling of how it is done in both in one place and then reconcile as we go. 

I'm imaging in Web that this stuff is statically handled as exception through a series of components. I would prefer to not do that on Mobile, and have an opinion that making simulated Groups (Contexts) for these keeps the UI code just UI. That aside, all the constant data related to these special groups can and I think should be shared, and I think the construction of these groups as valid static groups that match enough what comes from our API for mobile, is a good move, which I am happy to keep in this same shared module just-in-case it gets adopted in web. 

All of this should be platform non-specific.

Am I making sense, do you have differences, and do you know where to go from here?

lorenjohnson: I think the naming issue here is either with calling this a -Presenter or with Context vs Group. I think that is what underlies the confusion here and this is a good time to clear that up. That is what feels like where we're actually getting stuck, while talking beyond it.

My vote is to continue calling them Group until the dust settles and we refactor everything, perhaps backend included to switch to Context. In the mean time a Context is to me a ContextGroup which inherits most all of the attributes of a Group and extends it and perhaps modifies it in some cases. (For coherency I would go as far as naming the new ContextNav ContextGroupNav, but imagine I'm alone in that and I don't care as much at that level.) 

We should decide on something and then I suspect it will be easier to think about from thre.

apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
lorenjohnson: But that will be put in the presenter, right? So you will end up with a transformed widget that will have widget.inferredType? 

Also, on the presenter should resolve iconName. Even if you still need this special WidgetIcon component for other reasons, that logic is going to be used in Mobile as well.

apps/web/src/routes/AllView/AllView.jsx
```diff
@@ -93,7 +93,7 @@ export default function AllViews () {
 
   const widgetsSorted = useMemo(() => {
```
lorenjohnson, line 94: Is this the default ordering always? If so can we make the backend return them in this order by default, and if not lets create a useWidgetCollection hook that applies this default whenever we're using a widget collection.

apps/web/src/routes/AllView/AllView.jsx
```diff
@@ -3,8 +3,8 @@ import { useSelector, useDispatch } from 'react-redux'
 import { useTranslation } from 'react-i18next'
 import { useLocation, useNavigate, useParams } from 'react-router-dom'
 import { capitalize } from 'lodash'
-import { widgetUrl, widgetTitleResolver, widgetTypeResolver, isValidHomeWidget, humanReadableTypes, widgetIsValidChild } from 'util/contextWidgets'
-import { addQuerystringToPath, baseUrl } from 'util/navigation'
+import { ContextWidgetPresenter, widgetIsValidChild, humanReadableTypes, isValidHomeWidget, widgetTypeInferrer } from '@hylo/shared/src/ContextWidgetPresenter'
```
lorenjohnson: isValidHomeWidget, inferredType, and isValidChild seem to be like they should be resolved in the Presenter

apps/web/src/routes/AllView/AllView.jsx
```diff
@@ -102,13 +102,13 @@ export default function AllViews () {
     return widgetsSorted.map(widget => {
       const title = widget.title
       const url = widgetUrl({ widget, rootPath, groupSlug: routeParams.groupSlug, context: 'group' })
-      const type = widgetTypeResolver({ widget })
+      const type = widgetTypeInferrer({ widget })
       const capitalizedType = type.charAt(0).toUpperCase() + type.slice(1)
```
lorenjohnson, line 106: Resolve in presenter as typeTitle and same below viewTitle... Then user here as widget.typeTitle and widget.viewTitle...

apps/web/src/routes/AllView/AllView.jsx
```diff
@@ -93,7 +93,7 @@ export default function AllViews () {
 
   const widgetsSorted = useMemo(() => {
     return visibleWidgets.map(widget => {
-      return { ...widget, title: widgetTitleResolver({ widget, t }) }
+      return ContextWidgetPresenter({ widget, t })
```
lorenjohnson, line 96: Can the Context WidgetPresenter grab the translation `t` object itself?

apps/web/src/routes/AllView/AllView.jsx
```diff
@@ -102,13 +102,13 @@ export default function AllViews () {
     return widgetsSorted.map(widget => {
       const title = widget.title
       const url = widgetUrl({ widget, rootPath, groupSlug: routeParams.groupSlug, context: 'group' })
```
lorenjohnson, line 104: widget.url({ rootPath, groupSlug: routeParams.groupSlug, context: 'group' }) or if it is predictable from widget attributes and a route object then simply widget.url(route) ? ...

apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -64,7 +64,7 @@ import ThreadList from 'routes/Messages/ThreadList'
 import UserSettings from 'routes/UserSettings'
 import { GROUP_TYPES } from 'store/models/Group'
 import classes from './AuthLayoutRouter.module.scss'
-import { findHomeView } from 'util/contextWidgets'
+import { findHomeView } from '@hylo/shared/src/ContextWidgetPresenter'
```
lorenjohnson, line 67: Not clear on what its doing, but this on the surface doesn't look like something I'd not put in the Presenter. Without understanding it better, and assuming it belongs in the ContextWidgetPresenter module as a method outside of the default export of the transformer function then I would name it fully as findHomeWidgetView or something a bit more clearly designating its type amongst the options? What is a view by the way?

packages/shared/src/ContextWidgetPresenter.js
lorenjohnson: Gotcha. To me this would still be more clear and highlight the weird case(s) more which I think is a good thing:

```
return (
  widget?.type ||
  widget?.view ||
  (widget?.viewGroup &&  'viewGroup') ||
  widget?.viewPost ||
  // ...etc, 'customView' 
)
```

packages/shared/src/ContextWidgetPresenter.js
lorenjohnson: Looking now at this I would definitely put this as a `widget.isValidChild(parentWidget)` method on the return of `ContextWigetPresenter(widget)`.

apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
thomasgwatson: just resolving iconName isn't enough.

I'm using a separate 'icon' resolution component, because it actually might return several categories of components; Avatars, Lucide icons or the prior icon

apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.js
thomasgwatson: This is used to determine the default view for a group; so its more of a GroupPresenter thing.

apps/web/src/routes/AllView/AllView.jsx
thomasgwatson: There are two ordering currently. This one and orderContextWidgetsForContextMenu.

In time, this one might be split into categories and changed again, depending on the whims of Product/Design.

Because we are living in the reduxORM/Urql split existence, I've not married up how widgetCollections are handled. Leaving that for when we bring urql everywhere

packages/shared/src/ContextMenuPresenter.js
thomasgwatson: re handling ContextGroups on web vs mobile:
Mainly I see this falling out of the ability to use url state directly in web and not so much in Mobile. So I'd expect that difference to continue

lorenjohnson: Can you say more about "URL state"? I am not following entirely, an example might help.

thomasgwatson: just that state can be constantly accessed from the path of the url `/:context/:optionalContextId/:view/:optionalViewId` etc. /groups/building-hylo/members/45

apps/web/src/routes/AllView/AllView.jsx
lorenjohnson: But you can toRefArray() this and have a plain JS object which you can do a plain Array.sort on. Please do that as this code will mostly work after conversion as well. 

Also, are there any methods on the ContextWidget ReduxORM model which we're relying on? I forgot to ask. 

Consider creating useWidget and useWidgets hooks to make the component as dumb as possible and put the selecting, presenting/transforming (shall we just call these *Transformers?), and logic of a Widget in one place that offers back to the component just what it needs. 

If we have a hook that does that, which employs the presenter, then we can easily refactor any bits on conversion in one place. That is a practical reason, but it's also an example of why it's a solid architectural choice. 

We have the data presenter which is shared between front ends and platform agnostic. From there we have these hooks which have would probably be named the same on each platform, and would also likely do very similar things but one would be Redux/Web centric doing a redux selection, and the other on Mobile would do the URQL querying in the same place. If we set it up that way then the Mobile version will likely become shared once we convert.

What I had imagined, in not knowing how much was already on Web about all of this, is that we'd be developing this ContextWidgetPresenter on Mobile and then adapting back to Web. I assumed that for no good reason other than is how my process has been so far. It is I think clarifying though as the URQL is the future (echo echo echo)... Just noting this. I don't care as much about whether this is fully shared right away than that we're not setting ourselves on the right course and have a good enough shared sense of what that is and why.

apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.js
lorenjohnson: yep, that makes sense to me and we'll need this in mobile. so hold that though and once i push down the shared stuff from mobile we'll sort this out.

lorenjohnson:
You have merged:) I am fine with that, as long as we can we can finish a couple convos here through a paired session once the presenter has to be employed in Mobile.

PR #164, opened 2025-01-27
chore(mobile): URQL schema awareness and component and resolver updates for partial results

lorenjohnson:
This one does a few things that are related to where I was focused but may seem a bit unrelated without the context:

- Slays the dragon. `makePaginationResolver` handles both our cursor/limit and offset/first style of pagination, and I think is at this point doing a good job. There may be tuning here and there in the future, but am feeling actually fairly confident this one is pretty worked out and will serve us for some time to come as long as we keep doing our pagination in the pattern we do now. It is a short bit of code but it was so unclear how to get there... 
- Creates a createComment graphcache updater that, beyond making comment creation update nicely without re-queries, clears-up updater patterns for most other things where we need it. This was also really opaque and earlier attempts didn't go well. Now we're there (AND--funnily 😅, the createComment updater may have just broke this, but should be easy to fix. Will be left as is for this PR, and I'll be looping back)
- Some added tolerance for null results in some Mobile components to make Schema Awareness in URQL work. 
- Schema Awareness is setup for the URQL client but now disabled by default, as so far I've seen it causing more trouble than helping in our current Mobile setup. That may chance, and it's ready to turn on and try anytime. The URQL client has always been configured in awareness of the Web and it should just drop-in, and schema awareness may work wonders there where it is otherwise not a great fit for Mobile because of the rendering of FlatLists and such. TBD.
- Adds querySetId generation to the backend, which is not currently being utilized, and may never be. I have left a clear comment in the area where it is happening and may pull it out in coming weeks. I believe it to be harmless as it is, and if we want to wage any experiments with any graphcache reconciliation that may be helpful in coming weeks we simply need to add an `ID: String` to (most) any *QuerySet and there is an ID there waiting for us. 
- Turns off default dev behaviour of subscriptions events publishing also to their creator, and adds a request header option to enable this per subscription. It is handy, but messes with testing/configuring cache reconciliation as the subscription events were just updating things I was other trying to confirm got updated the normal way through the mutation (we don't really want to rely on subscriptions for that).
- I had to merge-by-hand the latest mutation you added on the Backend @tibetsprague as it went down to dev before this branch which split graphql/api/index.js into makeSchema.js. It was straight forward and I wouldn't expect it to have broken anything. 
- Another small thing or two...


apps/mobile/src/urql-shared/updates/index.js
```diff
@@ -4,73 +4,73 @@ import meCheckAuthQuery from 'graphql/queries/meCheckAuthQuery'
 
 export default {
   Mutation: {
-    clearModerationAction: (result, args, cache, info) => {
-      if (result[info.fieldName].success) {
-        cache.invalidate({ __typename: 'ModerationAction', id: args.moderationActionId })
-      }
-    },
+    // clearModerationAction: (result, args, cache, info) => {
```
lorenjohnson, line 7: Some of these updaters can go away for good, some will come back as proper atomic cache updates like createComment, others will still happily invalidate. Right now disabled so they can be re-worked and tested one-by-one outside the scope of this PR.

apps/backend/lib/graphql-bookshelf-bridge/fetcher.js
```diff
@@ -124,20 +124,35 @@ export default class Fetcher {
...
+    const instances = await this.loaders.relations.load({ relation, method })
+
+    // N.B. this caching doesn't take into account data added by withPivot
+    instances.forEach(x => loader.clear(x.id).prime(x.id, x))
+
+    const models = await loader.loadMany(instances.map('id'))
+
+    if (!fetchOpts.querySet) return models
+
+    // NOTE: querySetId generation, which may prove useful for graphql cache reconciliation in the
```
lorenjohnson, line 137: Adds the querySetId. As mentioned in PR description, I may remove this soon though harmless as it is.

apps/backend/lib/graphql-bookshelf-bridge/fetcher.js
```diff
@@ -124,20 +124,35 @@ export default class Fetcher {
     return matches[0][0]
   }
 
-  _loadMany (relation, { method, loader, tap, fetchOpts }) {
-    return this.loaders.relations.load({relation, method})
-    .tap(tap)
```
lorenjohnson, line 129: This `.tap` run was removed as I researched it and found nowhere where it was any longer used or ran. As a reminder, because I also had to look it up: .tap is a Bluebird Promise chain method that simply runs the passed function while passing along the results of the promise from before it to the next thing. In async/await language it is just running a function between two other functions. I would have just put it in here, but it was so clear that it wasn't in use I decided to make it go away.

apps/backend/.env.example
```diff
@@ -42,6 +42,12 @@ SLACK_APP_CLIENT_SECRET=
 # UPLOADER_HOST=
 UPLOADER_PATH_PREFIX=
 
+# See https://sailsjs.com/documentation/concepts/logging for options.
```
lorenjohnson, line 45: The sails logging setting of "silent" is super helpful for any debug logging other than query debugging in the Backend. Highly recommended. There are defaults on both of these in their config files so nothing new or bad happens if you don't set them in your .env.

apps/mobile/src/urql-shared/resolvers/makePaginationResolver.js
lorenjohnson, line 1: Very happy with this and the next files. I finally feel most of the way to loving URQL. The implementation is entirely coherent and mostly just graphql patterns, but the docs are pretty vague at times. I think it gets easier from here. There are some things to iron out in the UI, but at the actual URQL/graphql layer things are working quite well.

lorenjohnson:
@thomasgwatson I will be re-write or replacing some or all the URQL update queries soon, so don't panic if some of them broke. I didn't test yet.

apps/mobile/src/urql-shared/updates/index.js
```diff
@@ -107,10 +107,12 @@ export default {
         }
         case 'MessageThread': {
           console.log('!!!! updates TODO: new MessageThread. Increment Messages tab badge', result, args)
+          cache.invalidate({ __typename: 'MessageThread', id: update?.messageThread?.id })
```
lorenjohnson, line 110: Enables proper live updating to MessageThreads and Messages in he background, and working well now that the pagination caching issue is cleared-up.

apps/backend/lib/graphql-bookshelf-bridge/fetcher.js
tibetsprague: id love to move fully to async await and get rid of bluebird... in time

apps/mobile/src/components/StreamList/StreamList.js
```diff
@@ -206,7 +214,12 @@ export default function StreamList (props) {
             )}
             {streamType === 'proposal' && (
               <View style={[styles.listControls]}>
-                <ListControl selected={streamType} onChange={() => navigation.navigate('Decisions', { streamType: 'moderation', initial: false, options: { title: 'Moderation' } })} options={DECISIONS_OPTIONS} />
+                <ListControl
+                  selected={streamType}
+                  onChange={() => (
+                    navigation.navigate('Decisions', { streamType: 'moderation', initial: false, options: { title: 'Moderation' } })
+                  )} options={DECISIONS_OPTIONS}
```
tibetsprague, line 221: options should be on its own row

apps/mobile/src/urql-shared/resolvers/makePaginationResolver.js
```diff
@@ -0,0 +1,108 @@
...
+// the full set of QuerySet.items available from the incoming query results + the cache, and ordered
+// by the query designated arg for "order" which is assumed to be either "asc" or "desc". The ordering
+// and deduping may both be unnecessary, but are kept for as a safeguard.
+//
+// Test cases for offset pagination -- imagine:
+//
+// 1) Result should skip cache and query:
+//     posts(limit: 10 offset: 0 type: discussion, sortBy:updated)
+//
+// 2) Will query and then merge results with the results from step 1 (wash, repeat):
```
tibetsprague, line 20: i dont actually understand this. you are saying that even though the sortBy changed from updated to popular it will still merge the results, and so the frontend will have to sort that out?

tibetsprague (approved):

apps/mobile/src/urql-shared/resolvers/makePaginationResolver.js
lorenjohnson: no, all the args for the field. so in the case of the stream, all the variables we send into the query under `groups { posts(<here>) {} }` except the pagination-related make for the cache key match against for existing cache entries. in short, pages. if there are no pages already then don't process further just forward on without trying to add any (non)existing sets. you get me now?

apps/backend/lib/graphql-bookshelf-bridge/fetcher.js
lorenjohnson: for sure, you and I both

PR #176, opened 2025-01-30
Add sql migrations for the initial setup of context widgets, and for …

thomasgwatson:
…the syncing fo chats and custom views to the context menu. Still requires a step to reorder widgets based on a groups contents

### DIFF
--- a/apps/backend/api/models/Group.js
+++ b/apps/backend/api/models/Group.js
@@ -564,43 +564,42 @@ module.exports = bookshelf.Model.extend(merge({
     ))
   },
 
-  // This is idempotent
   async transitionToNewMenu (existingTrx) {
     const doWork = async (trx) => {
       // Get all widgets for this group
       const widgets = await ContextWidget.where({ group_id: this.id }).fetchAll({ transacting: trx })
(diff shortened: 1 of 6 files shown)


thomasgwatson:
I ran both of these commands in postico for my local db;

It did fix missing custom view widgets (which I had previously verified as being out of sync)

And the `setup` query resulted in no new widget records, suggesting that it was working fine.

You can run both on your local db and verify

thomasgwatson:
Note, it only runs on 'active' groups right now... so that might be an issue if someone seeks to reactive a group?? I don't know the mechanisms around that

lorenjohnson:
> Note, it only runs on 'active' groups right now... so that might be an issue if someone seeks to reactive a group?? I don't know the mechanisms around that

What is blocking also running it on inactive groups?

thomasgwatson:
Just trying to cut down on overall query time. Was doing this stuff on the Node side, and it was dreadfully slow. But I guess its not going to be that much more work to do inactive groups as well....

tibetsprague (approved):
ok im going to try running this from the branch and see how it goes in my environment

thomasgwatson:
@tibetsprague did this get merged elsewhere or should I merge it?

PR #177, opened 2025-01-30
update-widgets-based-on-group-content

thomasgwatson:
This still bulk runs Group Model methods on all groups but is a vastly constrained subset of the previous workload. So I think it will be fine to run as a migration.

Keeping it in a separate PR from the other related migrations, in case we want to do something different


thomasgwatson:
Looks good.

I've been trying hard in other parts of the implementation to ensure the order column is continuous; [1, 2, 3, 4, 5], never [1, 3, 4, 6, 7], etc. Toyed with a few ways of enforcing this (validator, 'settling' function in the db...) but haven't implemented anything.

Given this is the once-off migration, it would be good if we can add a query to settle each auto-add widget children ordering. That way each group will start from a clean ordering state. Other wise, we'll have to add a bunch of test cases for reordering mutations (which are all currently in the backend but ultimately I'd want to have in some front-end/mobile/backend shared folder)

Something like this: https://stackoverflow.com/questions/56500663/update-in-a-specific-order

I'm happy to do it if you are jumping on to something else

thomasgwatson:
"In other places where we do drag and drop reordering all I do is just send the new order of all the items and update all of of them to the new order."

This is actually what happens now. But because a single reordering mutation can impact various combinations of the parent order and/or different children ordering, it needs to have the exact same logic (reorderTree) on front-ends/backend to ensure the state is correctly represented on mutations. This is also why the params for that function don't directly accept something like `newOrder`; its more consistent to instead specify which widget to place the incoming reordered widget in front of (or addToEnd, when you just need to append to the end of an ordering).

I can give a variation of that stackoverflow a whirl

thomasgwatson:
Ok, added that last order settling, and tested it out on my db. Works good.

tibetsprague:
so i ran into a bug with the new migration because it runs on inactive groups, whereas the migration to setup the original context widgets including the auto add widget runs only on active groups. so we need to change one or the other. seems like probably a good idea to add widgets for inactive groups too, in case they get reactivated? ill let you decide which to change while i sleep :)

tibetsprague:
ok i went ahead and just made the new migration only work with active groups

thomasgwatson:
ahh fuck.. yeah that whole active/inactive thing was probably unnecessary

PR #191, opened 2025-02-02
Handles Heroku postbuild logic in a script, addressing slug size and overbuilding

lorenjohnson:
- Fixes slug size issue on Hylo-node deploys by re-running yarn install as `yarn workspaces focus backend --production` as a postbuild step in combination with turning on `YARN2_SKIP_PRUNING` via env var. Brings slug size back down to 130M from 503M.
- Adds a now required `YARN2_WORKSPACE_PATH=apps/web` to the hylo-evo-staging instance on Heroku. Was there a reason we didn't do this previously @tibetsprague? It seems like the right thing to do. This configuration does work currently deployed from this branch and postbuild script, and when combined with setting `YARN2_SKIP_PRUNING=true` the web bundle size goes down to 203M from the (also near limit) of 489M. 
- Now we only builds @hylo/shared on a backend deploy, and web (vite) on a web deploy
- Renames the `build:cjs` task in @hylo/shared to simply `build` as there is no other build output option

This branch is currently deployed successfully on both the staging backend and frontend.

‼️ UPON APPROVAL BEFORE MERGING:

- Add  `YARN2_WORKSPACE_PATH=apps/web` to hylo-evo production
- Add `YARN2_SKIP_PRUNING=true` to BOTH hylo-evo and hylo-node production


tibetsprague:
We are not using the [heroku-buildpack-yarn-workspace](https://github.com/loggedltd/heroku-buildpack-yarn-workspace) buildpack on the front end so YARN2_WORKSPACE_PATH won't mean or do anything. Right now the deployment of the front-end is super hacky, because it's the only way I could it working with vite. We are basically just deploying the root monorepo, then in the root package.json you'll see `heroku-postbuild: "yarn workspace web build && yarn workspace @hylo/shared build:cjs"` which builds the front-end (and shared) and then in the root Procfile `web: node apps/web/src/server/index.js`. So basically the front-end deploy all happens through deploying the root monorepo. The backend handles things nicely with the above buildpack + the heroku-buildpack-multi-procfile to do the build and run from the backend app directory. I can't remember right now why exactly the front-end was breaking, I hope i have notes written down somewhere. If you feel called to give it another shot go for it, but i spent days on it and gave up.

tibetsprague (commented):
and maybe this already fixes it! if this is correctly deploying and running the backend and the front-end on staging right now then it looks good to me. i do feel like this might not actually be doing anything for the backend though since the yarn2 workspace buildpack goes into the backend directory when you deploy and does the build and postbuild and everything else from there. we could try removing that buildback and seeing if this works with the backend

lorenjohnson:
> and maybe this already fixes it! if this is correctly deploying and running the backend and the front-end on staging right now then it looks good to me. i do feel like this might not actually be doing anything for the backend though since the yarn2 workspace buildpack goes into the backend directory when you deploy and does the build and postbuild and everything else from there. we could try removing that buildback and seeing if this works with the backend

Yes, it is working and deployed as configured here. In looking at the build log it does actually pay attention to the `YARN2_SKIP_PRUNING` setting via the node buildpack + yarn recognition as per:  

```
-----> Pruning devDependencies
       Skipping because YARN2_SKIP_PRUNING is 'true'
```

I do think you're correct that the multi buildpack is not doing anything for hylo-node (looking at the code for the buildpack and the logging from the logging for the deply confirms this), and the env var `YARN2_WORKSPACE_PATH` is being re-used in this heroku-postbuild.sh I've added to basically do what multi buildpacks did but for web. I will try removing the multi buildpack now. 

One thing though, is there anything which remains that feels "hacky" to you about the hylo-evo deploy setup? It looks pretty straightforward to me with these changes. Basically we're treating the frontend and backend the same, and in the heroku-postbuild.sh doing the things specific to them (building shared for node, and building vite for web). All in all I think we have landed perhaps somewhat by happenstance, on the correct and best setup here for now. Which is my hope, because just wanting this groundwork solid and under our feet as I jump into a final push:)

lorenjohnson:
Ok, tested without [the yarn workspace ](https://github.com/loggedltd/heroku-buildpack-yarn-workspace) buildpack for hylo-node-staging and it builds fine but was crashing on run. Could look into it further, and I suspect we can just consolidate the redirection that buildpack does into the heroku-postinall.sh hook, but for not going to fix what isn't broken. I think things are pretty good here and both apps are happy deploying to staging after multiple tests. 

Going to go ahead and merge down, including adding the env vars to the production instances as described so things should be good to go there once we go live.

PR #196, opened 2025-02-03
All view port to mobile

thomasgwatson:


lorenjohnson (changes requested):
@thomasgwatson feels to me like things are headed in the right direction 👍  There is enough here that I still see needing to move around that I've put my response to this PR in another PR on top of this branch. See notes there and let me know where I can meet you.

PR #197, opened 2025-02-03
All view port to mobile -- response

lorenjohnson:
This is a refactor from what you have in your branch @thomasgwatson. Some simple tidy-up and fixes, but also several larger moves here which I do care a fair amount about us making or at least genuinely considering as I do think we need things better factored here to avoid some traps coming in the next steps.  I feel much less picky about what happens as we move up the layers from there. . Give it a look and see if you follow. I expect there is a thing or two here not quite right or lost in translation. 

What's happening here in this PR branch:

- Moves ContextWidgetPresenter into @hylo/presenters and, for now, collapsed ViewHelpers and ContextMenuHelpers into that file, propagating that change where it is imported both in Web and Mobile. There is more here to do, like as you suggested, what was in ViewHelpers and ContextMenuHelpers likely is moving more towards GroupPresenter or another new place. I am comfortable with them in this home for now while we sort that out, and until I move GroupPresenter into the presenters module (will do that probably today)
- Resolves `iconName`, `avatarUrl`, and `displayName` into the ContextWidgetPresenter and then uses these static values in the two WidgetIcon components which does greatly simplify them (nearly making them unnecessary)
-  ContextWidgetPresenter adds  a `_presented` flag to an already presented widget object and checks for that on run of ContextWidgetPresenter(unknownWidget) so we never try and double transform a widget. This allows us FOR NOW to still be lazy about deciding when and if we present a widget and don't need to worry about over doing it if we're not sure it was already presented. 
- For testing in Mobile All Views can now be accessed through the GroupNavigation menu as the last item of a Group
- Adds new shared `CustomViewPresenter` for later elaboration as needed, currently only holds the static `COMMON_VIEWS` object
- Refactors to use presented widget attribute instead of exporting the resolvers where found
- Keeps context and null safety by using `widget?.attribute` vs `const { attribute } = widget` in a few places
- Some other things ?

And there are still 3-4 or things I'd like to iron out here, but if we can meet in this version then I can deal with those things... Please let's talk if I'm losing you here. I would still really like to be pair on these pieces until we're both feeling settled, and then dive into the remaining react navigation stuff to get us fully into the context menu setup until a point you can set sail again into that work?


apps/mobile/src/navigation/AuthRootNavigator.js
```diff
@@ -4,8 +4,8 @@ import { createStackNavigator } from '@react-navigation/stack'
 import Intercom from '@intercom/intercom-react-native'
 import { LogLevel, OneSignal } from 'react-native-onesignal'
 import { gql, useMutation, useQuery, useSubscription } from 'urql'
-import i18n from '../../i18n'
 import mixpanel from 'services/mixpanel'
+import { useTranslation } from 'react-i18next'
```
lorenjohnson, line 8: Sorry, this was just a "while I'm in here" tidy-up. Of no relevance to this PR otherwise.

packages/presenters/CustomViewPresenter.js
```diff
@@ -1,54 +1,70 @@
+export default function CustomViewPresenter (customView, { t }) {
```
thomasgwatson, line 1: I think this is the only thing that I don't think makes sense.

We have had our classic 'views' behavior scattered across many components for some time, basically responsive to a string, such as 'events'. Derived, typically, from a url. Off the back of that, I simply represented those views via ContextWidgets with contextWidget.view, allowing a context widget to model something in the menu that points at one of these views.

Tibet then decided to consolidate the logic of these derived views, so we would have a more consistent use of them.

If some of the common views look CustomView shaped, its because they largely go into the same hole: the Stream component. But several of them are not shaped like that, which is also true of some CustomViews (externalLink).

So I don't think it makes sense to bundle the CommonViews into this file, if its scoped to the the current concept of a CustomView

lorenjohnson: Gotcha. Well I'll go a round today factoring this stuff into some sorta hook that holds all the ContextMenu static things and move it out of here.

PR #202, opened 2025-02-03
Mobile update sub handler and messages fixes

lorenjohnson:
- Adds a `makeAppendToPaginatedSetResolver` which finally makes handling the difficult update queries reliable and simple enough to wire-up  (as our nested QuerySets have been most of the difficulty as well as the things needing custom updaters)
- As a product of that our new "updates" and "comments" subscription are now well handled and making things "live" in Mobile. New MessageThreads update the badge count on the Messages tab icon, new Messages get streamed in in real time when you're in a Thread chatting with someone , and Comments stream in when you're in a Post Detail view and someone comments from afar. Most--but not all--of these things theoretically worked with Sockets, but I was never sure they were working. So far in this setup things do just seem to work. There is a TODO: URQL! remaining to add some retyr/reconnect handling to the subscription handler, otherwise this set of things feels wrapped.
- Santizes @hylo/urql/makeUrlqClient.js by making the subscription handler configurable and moving the mobileSubscriptionHandler.js back local to apps/mobile/src/urql/. This can rest now, we are ready to start converting Web with just dropping in the client and running whenever/if ever (yes we will) get there. 
- A lot of small refinements to the Messages experience along the way of getting it updating correctly. It is becoming nice to use, and I expect overall more reliable.
- Adds `markThreadRead(messageThreadId: ID)` mutation to use instead of `PostContorller#updateLastRead`, which was helpful to smoothing-out the graphcache reconciliation around unseen/unread.
- Removes lots of unused / recently deprecated stuff


thomasgwatson (approved):
lgtm

lorenjohnson:
Hey @thomasgwatson I need a re-review on this one. I didn't even realize I was still working away off this branch, but it was also the right place to complete this work. This PR now brings URQL update resolvers and subscriptions very close to done done, which results in things being much more "alive" in the app (MessageThreads, Messages, and Comments on a currently open post). Also Messages and Notifications should get lit up when new things come in. Very exciting update here for me. And sorry to have to send you back through it! Thank you!

packages/graphql/mutations/createMessageMutation.js
```diff
@@ -13,20 +13,21 @@ export default gql`
     $messageThreadId: String,
     $text: String
     $createdAt: Date
-    $firstMessages: Int = 1
+    # $firstMessages: Int = 1
```
thomasgwatson, line 16: Does this comment something out? I'm not familiar with the syntax?

packages/urql/updates/index.js
```diff
@@ -1,90 +1,129 @@
...
+import { get } from 'lodash/fp'
+import meQuery from '@hylo/graphql/queries/meQuery'
 import meCheckAuthQuery from '@hylo/graphql/queries/meCheckAuthQuery'
+import makeAppendToPaginatedSetResolver from './makeAppendToPaginatedSetResolver'
 
 export default {
   Mutation: {
-    addProposalVote: (result, args, cache, info) => { 
+    addProposalVote: (result, args, cache, info) => {
+      // TODO: URQL - This will result in Me being fully re-queried with every new skilled added, which may be fine...
```
thomasgwatson, line 9: Is this comment directed at a different mutation? Sounds like you are referring to skills, but this is about proposalVotes, which I would expect to impact a Post (and linked collections) but not Me/currentUser?

packages/urql/updates/index.js
```diff
@@ -1,90 +1,129 @@
...
+      // but probably can apply makeAppendToPaginatedSetResolver here as it is a QuerySet
       cache.invalidate({ __typename: 'Post', id: args.postId })
     },
+
     addSkill: (result, args, cache, info) => {
+      // TODO: URQL - This will result in Me being fully re-queried with every new skilled added, which may be fine...
```
thomasgwatson, line 15: Ok, I see this one here, maybe the above comment still applies in a related way?

packages/urql/updates/index.js
```diff
@@ -1,90 +1,129 @@
...
         cache.invalidate({ __typename: 'ModerationAction', id: args.moderationActionId })
       }
     },
-    createComment,
+
+    createComment: (result, args, cache, info) => {
+      const parentId = args?.data?.parentCommentId || args?.data?.postId
+      const parentType = args?.data?.parentCommentId ? 'Comment' : 'Post'
+      const fieldName = args?.data?.parentCommentId ? 'childComments' : 'comments'
+      makeAppendToPaginatedSetResolver({
```
thomasgwatson, line 30: So this is the general pattern for working with our abundance of nexted querySets?

packages/urql/updates/index.js
```diff
@@ -1,90 +1,129 @@
...
     },
+
     removeProposalVote: (result, args, cache, info) => { 
       cache.invalidate({ __typename: 'Post', id: args.postId })
     },
+
     pinPost: (result, args, cache, info) => {
       if (result[info.fieldName].success) {
-        // Any Post invalidation will result in the full Group/Stream query being re-fetched.
+        // Note: Any Post invalidation will result in the full Group/Stream query being re-fetched.
```
thomasgwatson, line 105: Acknowledged

packages/urql/updates/index.js
```diff
@@ -93,31 +132,44 @@ export default {
...
+          cache.updateQuery({ query: meQuery }, ({ me }) => {
+            if (!me) return null
+            return { me: { ...me, newNotificationCount: me.newNotificationCount + 1 } }
+          })
+          return
         }
+
         default: {
-          console.log('!!! Unhandled update from updates subscription', result)
+          console.log('ℹ️ Unhandled update from updates subscription', result, args)
```
thomasgwatson, line 172: what do you think: should we use a console.warn for something like this?

packages/urql/updates/makeAppendToPaginatedSetResolver.js
```diff
@@ -0,0 +1,110 @@
+import { PAGINATION_ARGS } from '../resolvers/makePaginationResolver'
+
+/*
```
thomasgwatson, line 3: Love this, thank you

thomasgwatson (approved):
lgtm

packages/urql/updates/index.js
lorenjohnson: Ah yes, sorry. Was copy/pasted. Yes, when this one invalidates it will almost certainly trigger a re-query of the entire current group { posts } set. But it's usually fine and isn't as bad as it sounds as it happens in the background. Just wanted to point it out. And we will get these more refined soon, I finally am feeling confident of that, and there will be better examples in here to reference.

packages/urql/updates/index.js
lorenjohnson: Yeah...

packages/urql/updates/index.js
lorenjohnson: Exactly, but only where needed, it isn't always/usually.

packages/urql/updates/index.js
lorenjohnson: Sorry if I was repetitive :) Just putting these in there for later tuning/refactoring. As it is the perf on this stuff is not a big deal.

packages/urql/updates/index.js
lorenjohnson: No, I don't think it needs anything more than this. Nothing bad really happened if we hit this, and we'd likely only see it if we were debugging and in the process of adding a new updates type.

packages/urql/updates/makeAppendToPaginatedSetResolver.js
lorenjohnson: Thank you for seeing me :) OMG I sweat blood to get to to that, and it indeed solves most of the URQL woes to this point. Both in understanding how to get there, and in its actual utility.

packages/urql/updates/index.js
thomasgwatson: no no, just wanted to note I read this! I just wanted an emoji react really

PR #212, opened 2025-02-04
Stream+StreamList Consolidation in Mobile ++

lorenjohnson:
_🚧 NOTE: This branch is based off of dev + #197, so that needs to get settled and go down first_

- Refactors Mobile `Stream` to be more clear, and to now subsume `StreamList` (finally). Includes refactors of the related components of PostRow, StreamHeader, etc. Hopefully this will make things a fair bit easier to approach with upcoming Widget/View elaborations. 
- Moves `COMMON_VIEWS` into `ContextWidgetPresenter`--for now. 
- Removes most of what was in `GroupPresenter`: On closer look what was in there was just a transcription and merge of the old ReduxORM Group model + `presentGroup` which was helpful during the URQL transition. Basically all of the data massaging it was doing is now not being used or particularly helpful with how things are now in URQL on Mobile. This will leave open that Presenter for things like group.contextWidgets which is now doing something in there. 
- Renames `isContextGroup` to `isContextGroupSlug` because that naming was always misleading, and now we're using it more places.
- Removes some unused code, and otherwise straightens out some places around like Group Welcome, and others, where we were looking at group.agreements vs group.agreements.items, and the like...

...and some other things. This branch is one iteration of what I imagine will be 1-3 more of Mobile component clean-up which puts us in a more workable place to extend further from.


packages/presenters/ContextWidgetPresenter.js
```diff
@@ -1,8 +1,13 @@
 // TODO: Get "t" from current i18n instance so it doesn't need to be passed
 // import i18n from 'react-i18next'
-import { COMMON_VIEWS } from '@hylo/presenters/CustomViewPresenter'
 
-export default function ContextWidgetPresenter (widget, { t }) {
+// t / translate is currently set to a dummy function if not sent
+const tDummy = t => {
```
thomasgwatson, line 5: I'd rather be manually pushing t into presenters and being closer to the correct UI than this approach. 

I haven't push on this much yet, aside from running afoul of the rules of hooks.

thomasgwatson (approved):
approved with comment about the i18n stuff

packages/presenters/ContextWidgetPresenter.js
lorenjohnson: Absolutely! Yes good catch. This is very definitely a messy in-development bypass which definitely will be removed before live. I should have and still should make that much more clear here with a "TODO!" comment of a sort.

thomasgwatson:
have you slept?!

PR #213, opened 2025-02-05
Notifications-and-urlq-improvements

lorenjohnson:
🚧 NOTE: This branch is based off of #212 (which is atop #197), so those need to get settled and go down first. Review and notes/approvals before then will be helpful though to clearing the queue. I won't merge until we're all settle on what lies beneath. 

Some happy things in here 🙌🏼

- Makes URQL mobile subscription handler Single Shared SSE Connection (One stream for all subscriptions), vs making a connection for each subscription. Yay! This goes well with Mobile in particular.
- Adds exponential backoff retry/reconnect to the subscriptions handler
- Adds the ability to makeAppendToPaginatedSetResolver to append a mutation result into a base Query paginated set (i.e. notifications)
- Notifications form updates subscription get appended to notifications set now vs only incrementing unseenNotifications count
- Cleans-up some dusty code and makes some minor styling tweaks in Notifications list area

I don't expect to have to touch subscriptions stuff much more, and I think this is the first PR in like 10 which is actually adding more code than removing :) I guess we're turning the corner to building back up! (Widgets and tigers and bears oh my!)


thomasgwatson (approved):
lgtm

PR #214, opened 2025-02-05
merge filter, types in stream / query

KevinTriplett:
This merges the filter and types variables for clarity and consistency.

Is this acceptable as a PR?

### DIFF
--- a/apps/web/src/routes/Stream/Stream.js
+++ b/apps/web/src/routes/Stream/Stream.js
@@ -100,21 +100,13 @@ export default function Stream (props) {
   const childPostInclusion = querystringParams.c || defaultChildPostInclusion
   const timeframe = querystringParams.timeframe || 'future'
 
-  // TODO: merge this and getTypes
-  const determinePostTypeFilter = useCallback(() => {
-    if (view === 'projects') return 'project'
-    if (view === 'decisions') return 'proposal'
-    if (view === 'events') return 'event'
-    return querystringParams.t || defaultPostType
-  }, [querystringParams, defaultPostType, view])
+  const getPostTypes = useCallback(() => {
+    return (customView?.type === 'stream')
+      ? customView?.postTypes
+      : systemView?.postTypes || querystringParams.t || defaultPostType || null
+  }, [customView, systemView])
 
-  const postTypeFilter = determinePostTypeFilter()
-
-  const getTypes = useCallback(({ customView }) => {
-    if (customView?.type === 'stream') return customView?.postTypes
-    if (systemView) return systemView?.postTypes
-    return null
-  }, [systemView])
+  const postTypes = getPostTypes()
 
   const topics = topic ? [topic.id] : customView?.type === 'stream' ? customView?.topics?.toModelArray().map(t => t.id) : []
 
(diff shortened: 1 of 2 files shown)


thomasgwatson:
Defo suitable as a PR. The code looks fine, and... i just don't know if there is something sneaky that we are missing by making this change. I'm assuming you made the changes and tested it out and nothing exploded, so should be ok right?

apps/web/src/store/reducers/queryResults.js
```diff
@@ -313,7 +313,6 @@ export const queryParamWhitelist = [
   'collectionToFilterOut',
   'createdBy',
   'farmQuery',
-  'filter',
```
tibetsprague, line 316: whoa this is a dangerous change, have you tested all post queries to make sure that the query cache works for them and new posts appear in them correctly?

apps/web/src/store/reducers/queryResults.js
```diff
@@ -328,8 +327,7 @@ export const queryParamWhitelist = [
   'slug',
   'sortBy',
   'topic',
-  'type', // TODO: why do we have type & filter? should only need one
```
tibetsprague, line 331: also this is dangerous and needs lots of testing

apps/web/src/store/reducers/queryResults.js
```diff
@@ -328,8 +327,7 @@ export const queryParamWhitelist = [
...
-  // 'types', TODO: add types?
+  'types',
```
tibetsprague, line 330: also this is dangerous and needs lots of testing

apps/web/src/routes/Stream/Stream.js
```diff
@@ -308,12 +299,12 @@ export default function Stream (props) {
             newPost={newPost}
             querystringParams={querystringParams}
             routeParams={routeParams}
-            type={postTypeFilter}
```
tibetsprague, line 311: this used to be a single type, now its an array but you didnt change the code that works with this prop, are you sure its working?

apps/web/src/routes/Stream/Stream.js
```diff
@@ -308,12 +299,12 @@ export default function Stream (props) {
...
+            type={postTypes}
           />
         )}
         <ViewControls
           routeParams={routeParams} view={view} customPostTypes={customView?.type === 'stream' ? customView?.postTypes : null} customViewType={customView?.type}
-          postTypeFilter={postTypeFilter} sortBy={sortBy} viewMode={viewMode} searchValue={search}
+          postTypeFilter={postTypes} sortBy={sortBy} viewMode={viewMode} searchValue={search}
```
tibetsprague, line 307: this used to be a single type, now its an array but you didnt change the code that works with this prop, are you sure its working?

tibetsprague (changes requested):
some dangerous stuff here, I dont think you want the queryResults changes

apps/web/src/store/reducers/queryResults.js
KevinTriplett: Ah, you're right. When I create a new post, it doesn't show up. Hmm ... okay, let's abandon this one.

apps/web/src/routes/Stream/Stream.js
KevinTriplett: Ooh, you're right -- Updated to take the first array element

apps/web/src/routes/Stream/Stream.js
KevinTriplett: Updated to take the first array element

PR #215, opened 2025-02-05
return to prev location on post modal close

KevinTriplett:
Addresses [this comment](https://github.com/Hylozoic/hylo/commit/092d0fe548efa80dd605c5a6a17310674fbcc1e6#diff-48894935b7df20376b9be76d5f6165169d86ae2faacda6cd702f56a5385eb2d9R24) 

When editing a post, the modal cannot close because there is no "create" in the URL. So the URL doesn't change and therefore the modal doesn't close.

This is because the post edit function uses the CreateModal component. We can have two modals, one for create and one for edit, I suppose, but this PR seems to work.

**Is there an edge case that this PR is not addressing?**

### DIFF
--- a/apps/web/src/components/CreateModal/CreateModal.js
+++ b/apps/web/src/components/CreateModal/CreateModal.js
@@ -1,7 +1,9 @@
 import React, { useRef, useState } from 'react'
+import { useSelector } from 'react-redux'
 import { Route, Routes, useNavigate, useLocation } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
 import { CSSTransition } from 'react-transition-group'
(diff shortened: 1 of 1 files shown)


tibetsprague (changes requested):
hah, this literally reverts a change I just made which was causing other errors. but I forgot about the edit modal. this will cause other issues, so let me think about it. I think we just need to have it remove /create/* and /edit from the URL

KevinTriplett:
Yah, I saw that, which is why I made it a PR instead of pushing it to dev.

Removing the 'edit' part of the URL sends us back to the post detail, which may not be where we came from in the base of a PostCard. But it's much better than the modal not closing, lol.

Issue #183, opened 2025-01-30
Opening email link to chat leads to infinite loading in chatroom because of query-string


thomasgwatson:
I open `https://www.hylo.com/groups/hylo-stewards/topics/general/?postId=82634?ctt=post_email&cti=30623`

Opens to the chatroom but then stalls with infinite load

I remove:
?ctt=post_email&cti=30623

And now it loads

PR #218, opened 2025-02-06
Fix the notification URL to chat posts

tibetsprague:
Clean up notification post URLs a lot. Move all of it to Frontend.Route.post

Note, these URLs will have to be updated on mobile

### DIFF
--- a/apps/backend/api/models/Notification.js
+++ b/apps/backend/api/models/Notification.js
@@ -69,13 +69,6 @@ module.exports = bookshelf.Model.extend({
     return this.reader().get('settings')?.locale || this.actor().getLocale()
   },
 
-  postUrlHelper: function ({ post, group, isPublic = false, topic, reader }) {
-    if (post.get('type') === Post.Type.CHAT) {
-      return Frontend.Route.chatPostForMobile(post, group, topic)
-    }
-    return Frontend.Route.post(post, group, isPublic, topic)
-  },
-
   send: async function () {
     if (await this.shouldBeBlocked()) {
       this.destroy()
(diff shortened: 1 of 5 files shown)


apps/backend/api/models/Notification.js
```diff
@@ -434,7 +427,7 @@ module.exports = bookshelf.Model.extend({
         post_title: decode(post.title() || ''),
         post_topic: firstTag,
         post_type: post.get('type'),
-        post_url: Frontend.Route.tokenLogin(reader, token, this.postUrlHelper({ post, isPublic: false, topic: firstTag, group }) + '?ctt=post_email&cti=' + reader.id),
+        post_url: Frontend.Route.tokenLogin(reader, token, Frontend.Route.post(post, group, 'ctt=post_email&cti=' + reader.id)),
```
thomasgwatson, line 430: Do you think this fixes https://github.com/Hylozoic/hylo/issues/183 ?

apps/backend/api/services/Frontend.js
```diff
@@ -145,16 +137,22 @@ module.exports = {
...
 
-      if (isPublic) {
+      if (!group) {
         groupUrl = '/public'
       } else if (!isEmpty(groupSlug)) {
-        groupUrl = `/groups/${groupSlug}` + (topic ? `/topics/${topic}` : '')
+        const tags = post.relations.tags
+        const firstTopic = tags && tags.first()?.get('name')
+        if (firstTopic && (post.get('type') === Post.Type.CHAT || group.hasChatFor(firstTopic))) {
+          return url(`/groups/${groupSlug}/chat/${firstTopic}?postId=${post.id}&${extraParams}`)
```
thomasgwatson, line 150: Is the main update needed for mobile that it needs to track `/chat/` related urls?

apps/backend/api/models/Notification.js
tibetsprague: hmm, i doubt it, that sounds like a front end issue. but its not happening for me

apps/backend/api/services/Frontend.js
tibetsprague: yup

tibetsprague:
we need this on staging so im going to merge now

apps/backend/api/services/Frontend.js
thomasgwatson: Ok, I have that in a current branch

PR #220, opened 2025-02-06
Context menu cometh

thomasgwatson:
@lorenjohnson Ended up having to go shopping so didn't rejig the GroupMenuHeader to actually show yet but I'll stop here and you can hit the navigation stuff with hammers from this point, if you so desire.


apps/mobile/src/hooks/usePostActionSheet.js
```diff
@@ -57,14 +57,14 @@ export default function usePostActionSheet ({
   const [, pinPost] = useMutation(pinPostMutation)
   const { showHyloActionSheet } = useHyloActionSheet()
   const mixpanelTrack = useMixpanelTrack()
-  const [{ currentGroup }] = useCurrentGroup()
+  const [{ currentGroup, isContextGroupSlug }] = useCurrentGroup()
```
thomasgwatson, line 60: I am thinking.... @lorenjohnson 

`isContextGroupSlug` should refer to the function; and we should change the boolean that pops out of useCurrentGroup as `isContextGroup`.

Two reasons; 1) using different symbols for related but different entities 2) `isContextGroup` is probably more semantically aligned when coming out of useCurrentGroup, since its all references to the current group

lorenjohnson: yes for sure, go at it. it was just a bulk rename. definitely should be isContextGroup here, however my feeling here is that we should just promote that up to the now almost blank GroupPresenter as an isStatic property, or something similar that will translate coherently also in Web. In Mobile, "Context Group" meant a group that was statically defined and only there to create a context. I think that groups to us have finally entirely become understood as contexts and so the only thing left to delineate these previously "context groups" in mobile would be i think most sensibly staticGroups or systemGroups, so isStatic is my vote...

lorenjohnson:
@thomasgwatson I did a fairly quick run through of everything that is here just now, and had various comments, but wasn't sure what I was reviewing yet. Like at what stage are we and how through should I be in reviewing now? Do you be up for editing the description to make we have some notes there about what is done, known issues as is, and what remains to be done?

apps/mobile/src/navigation/linking/index.js
```diff
@@ -36,12 +36,16 @@ export const routingConfig = {
 
   '/hylo-editor':                                            `${AUTH_ROOT_SCREEN_NAME}/HyloEditor`,
 
-  // context group routes (/all, /public)
-  '/:groupSlug(all|public)':                                 `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab`,
+  // context group routes (/all, /public, /my)
```
lorenjohnson, line 39: @thomasgwatson you had a typo in the screen path here that I fixed, but I also removed your comment question about whether the screen path should "just include /Stream?" (paraphrase) The answer to that is, sure, if it is what is making sense now. I had let "Home Tab" handle what to do with it before (so basically sending it to the default which would have been Group Navigation which would have switched to the context and then forwarded to Stream. It will be a little different now, so I don't know the answer. But if you want /all and /public to go directly to the Steram for now, then this is the correct configuration.

apps/mobile/src/screens/GroupWelcomeFlow/GroupWelcomeLanding.js
```diff
@@ -54,6 +54,8 @@ export default function GroupWelcomeLanding ({ route }) {
   const [, updateMembershipSettings] = useMutation(updateMembershipMutation)
   const navigation = useNavigation()
   const [{ currentUser }] = useCurrentUser()
+  // TODO redesign: there are instances where this can be rendered and then the groupSlug is changed to a contextGroup and this explodes. Need to address this
+  // TODO redesign: this also gets weird when you switch groups; the selected group renders the Group Welcome, even if you've already done that for that group
```
lorenjohnson, line 58: Let's work through these cases before we call this branch done. The reason I say that is I think that in getting to the bottom of that issue will send us through getting to coherency on our Static / "Context Groups", change group routing, all in combo with the new patterns coming in, and that is something I think we should feel pretty settled about in this branch. That pattern is what I think we need to feel really clear and happy with before we move from this branch. 

_To the specific here I suspect that part of what is happening is now that we're relying on a presented group's group.isContextGroup we may want to add that property statically to each of the 3 static groups too with a value of true. Something like that, or otherwise, but again the goal here is to get to a clear sense of what is working, how, and why and not leave any open questions about that before moving on._

packages/hooks/useGetContextWidgetChildren.js
```diff
@@ -0,0 +1,37 @@
+import { useState, useEffect } from 'react'
```
lorenjohnson, line 1: Is this in progress or going away? Can we finish it? Can we finish the `useGatherItems` conversion as part of this branch? Also, what if any (or all 🤞🏼) of this can be handled in the presenter?. I'm not yet seeing why not nor why this hook would need a loading start. What I see is a a presented widget with childWidgets, and each child widget has a bound fetch method which can be called as needed when needed. Would love to pair on this one to make sure I'm getting the full requirement, and then make a recommendation from there.

apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,217 @@
...
+import Loading from 'components/Loading'
+
+export default function ContextMenu () {
+  const { myHome, groupSlug } = useRouteParams()
+  const [{ currentGroup, fetching }] = useCurrentGroup({ setToGroupSlug: groupSlug })
+  const currentGroupSlug = useCurrentGroupSlug()
+  const navigation = useNavigation()
+  const { t } = useTranslation()
+  const contextWidgets = currentGroup?.contextWidgets || []
+  const orderedWidgets = useMemo(() =>
```
lorenjohnson, line 29: Let's probably make this a useWidgets hook? The process of trying to do so, or answering to me why not will help us clarify the pattern a lot. I think this pattern is coming from Web, but I am a little confused in the Mobile setup by the use of myHome vs currentGroup. We may not be on Web, but I am fully ready and would prefer to entirely switch to the "Context" naming at this level and do away with the "Group" naming as I think it is obscuring things here. There may be "group" type contexts specific stuff that happens here and certainly below this level, but I think it should be handled semantically as a type of context more clearly. If my mental model is correct, the more we do that here, and I think we can with low effort, the more things will tighten up.

apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,217 @@
...
+      <ContextHeader
+        group={currentGroup}
+      />
+      <ContextWidgetList
+        contextWidgets={orderedWidgets}
+        groupSlug={currentGroup.slug}
+        rootPath={`/groups/${currentGroup.slug}`}
+      />
+      {/* Add the All Views widget at the bottom, matching web behavior */}
+      {(!isMyContext && !isPublic && !isAllContext) && (
```
lorenjohnson, line 51: Just make this `!currentGroup.isContextGroup` and do away with the `const`s above? I realize that there will likely be use of the differences between the three coming in here soon, but would rather keep that out until we see what emerges. We should be able to handle this simply by an attribute on a presented group, like group.type? OR whatever the right naming is for that as type may already be used for something else? Ideally we are switching here based on a presented group's attributes already resolved, with "type" being the most primitive.

apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,217 @@
...
+      {!widget.view && (
+        <View className='flex-row justify-between items-center content-center'>
+          <Text className='text-sm font-semibold text-foreground'>{title}</Text>
+        </View>
+      )}
+      <View className='flex flex-col justify-center items-center relative'>
+        {/* Need to add the SpecialTopElementRenderer here */}
+      </View>
+      {loading && <Text>{t('Loading...')}</Text>}
+      {listItems.length > 0 && listItems.map(item => <ListItemRenderer key={item.id} item={item} rootPath={rootPath} groupSlug={groupSlug} handleWidgetPress={handleWidgetPress} />)}
```
lorenjohnson, line 129: Please put a component on a new line and prop all on separate lines when it is s a long list (like ~ over 120 characters wide). This will be challenged in several cases, like I would put the Loading... component above also between parens and on a new line, but totally won't complain if you don't. Long lines like this though I have fairly consistently in the Mobile code base put on their own line, and long props lists get new lines. I'd like to keep doing that:

```
listItems.length > 0 && listItems.map(item => (
  <ListItemRenderer
      key={item.id}
      item={item}
      rootPath={rootPath}
      groupSlug={groupSlug}
      handleWidgetPress={handleWidgetPress}
    />
  ))}

File: apps/mobile/src/components/ContextMenu/ContextMenu.js [hunk 8]
```diff
@@ -0,0 +1,217 @@
...
+    </View>
+  )
+}
+
+function ListItemRenderer ({ item, rootPath, groupSlug, handleWidgetPress }) {
```
lorenjohnson, line 134: ListItem is not descriptive to me.

apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -0,0 +1,217 @@
...
+}
+
+function ContextHeader ({ group }) {
+  const { t } = useTranslation()
+
+  const isPublic = group?.slug === PUBLIC_CONTEXT_SLUG
+  const isMyContext = group?.slug === MY_CONTEXT_SLUG
+  const isAllContext = group?.slug === ALL_GROUPS_CONTEXT_SLUG
+  const isGroupContext = !isAllContext && !isMyContext && !isPublic
+  return null
```
lorenjohnson, line 201: You see what I mean about that isPublic, isMyContext, etc thing repeating everywhere. This is why we want this stuff on the presenter and to think it through at that level.

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -0,0 +1,137 @@
...
+        <TouchableOpacity 
+          onPress={() => {}} // TODO redesign: Needs to open some creation dialog...
+          style={styles.rowTouchable}
+          activeOpacity={0.7}
+        >
+          <View
+            className={`
+                bg-primary relative flex flex-col text-primary-foreground items-center justify-center
+                w-14 h-14 min-h-10 rounded-lg drop-shadow-md opacity-60 scale-90
+            `}
```
lorenjohnson, line 42: This is one way to handle the long className lists. I for now am mostly looking the other way and not going to worry about the inevitably very long lines here with the introduction of NativeWind, but wanted to highlight this option.

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -0,0 +1,137 @@
...
+import useCurrentUser from '@hylo/hooks/useCurrentUser'
+import useCurrentGroup from '@hylo/hooks/useCurrentGroup'
+import { PUBLIC_GROUP } from '@hylo/presenters/GroupPresenter'
+import useChangeToGroup from 'hooks/useChangeToGroup'
+
+export default function ContextSwitchMenu () {
+  const [{ currentUser }] = useCurrentUser()
+  const [{ currentGroup }] = useCurrentGroup()
+  const changeToGroup = useChangeToGroup()
+  const myGroups = [PUBLIC_GROUP].concat(sortBy('name', map(m => m.group, currentUser.memberships)))
```
lorenjohnson, line 15: Wondering if you know what order the memberships come in from the query? I assume not already sorted alphabetically by group name... (I also changed this to lodash methods for brevity/clarity here. I generally prefer lodash sortBy for things like this, but don't care a lot about that. Just wanted to remind you of options.)

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -0,0 +1,137 @@
...
+      style={styles.rowTouchable}
+      activeOpacity={0.7}
+    >
+      <View
+        className={[
+          'bg-primary relative flex flex-col items-center justify-center w-14 h-14 min-h-10 rounded-lg drop-shadow-md opacity-60 scale-90',
+          highlight && 'border-3 border-secondary opacity-100 scale-100',
+          badgeCount > 0 && 'border-3 border-accent opacity-100 scale-100',
+          className
+        ].filter(Boolean).join(' ')}
```
lorenjohnson, line 79: Please go ahead add clsx again and apply here and in similar cases. Didn't intend to veto it for good, was just trying to understand where we were headed in the earlier PR where it came up.

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -0,0 +1,137 @@
...
+          <View style={styles.badge}>
+            <Text style={styles.badgeText}>{newPostCount}</Text>
+          </View>
+        )}
+      </View>
+    </TouchableOpacity>
+  )
+}
+
+const styles = StyleSheet.create({
```
lorenjohnson, line 99: Let's talk at some point about what we keep do with NativeWind classNames and what we keep as React Native styles. I don't have the answer at the moment, and am not expecting us to know perfectly yet, but you may have a better sense than I already.

lorenjohnson (changes requested):
Ok, that is plenty of commentary for now. The roll-up is that I'd like us to have a further session about the Widget presenter, widget.childWidgets / items, and all of that until we come up with a more convincing pattern. I don't see a better time to finish those convos.

Also would like to strongly consider moving to Context language (vs Group or ContextGroup, etc) as appropriate but  consistently throughout this changeset in Mobile, even if we're not there yet on Web. I think it will help the clarity of thinking and implementation a lot from here if we settle that at this stage.

Other than that my code clean-ups and comments please take note of and push any further changes you wish in relation to them. Also, please let me know if I can help in anyway getting you on the road with linting again.

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
thomasgwatson: I don't know. Just following what Tibet set in the GlobalNav component on web

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
thomasgwatson: At a minimum:
- Because of the themeing setup, we more or less have to port everything impacting color over to nativewind.
Beyond that, I don't have a strong impulse of a need either way.

apps/mobile/src/screens/GroupWelcomeFlow/GroupWelcomeLanding.js
thomasgwatson: What I think is going on;

- Navigate to a group that you haven't gone through the welcome checks (or reset this by query in your local db)
- GroupWelcomeCheck (tucked into Chat and Stream) does a bunch of checks; if you fail them, it navigates to GroupWelcomeLanding (which is in the HomeNavigator stack but maybe should be a modal?).

Now, you can back out of the Welcome or open the menu while its open and go to another group
- but your navigation state still what it was; GroupWelcomeLanding is on top of the stack
- so now with the global group state changed, it tries to rerender that screen with the new group. This either breaks (because its a different context that doesn't have the required settings) or renders the welcome for a group you've already been welcomed too...

Previously this didn't really happen because we were also navigating to the Stream when we changed contexts/groups: https://github.com/Hylozoic/HyloReactNative/blob/dev/apps/mobile/src/hooks/useChangeToGroup.js

Sitting here pondering what to do about this...

We could wrap the 'check' logic in `GroupWelcomeCheck` into a hook and then share the hook between both `GroupWelcome` components, and offer different re-directs based on that.

packages/hooks/useGetContextWidgetChildren.js
thomasgwatson: The goal of this hook is to have a uniform interface for getting the children of a widget.

Often, its just their literal children. But some widgets have a 'dynamic' children.

At the moment, its just the 'members' widget. On web its literally just requesting the first four members but the goal is that it will request the four most recently active members. In the future, there will be others that dynamically request their children. The only likely one at the moment is 'Recent Activity', which will return posts.

Sure, we could refractor it. Does it need to be done on this PR? I'd rather address it in a targeted PR

apps/mobile/src/screens/GroupWelcomeFlow/GroupWelcomeLanding.js
lorenjohnson: My hunch is that this check has always most belonged in the useChangeToGroup setup, or directly in useCurrentGroup (setCurrentGroupSlug). The first is what always should be ran when changing between groups in the App, and the second (or somewhere within the interrelated hooks there) is always ran when we get the CurrentGroup, even on boot. So probably need to hook in there and have useCurrentGroup return another attribute like shouldWelcome or similar which then we check for wherever we want to make sure the GroupWelcome appears. Some options, pick whatever you think is the cleanest implementation.

apps/mobile/src/components/ContextMenu/ContextMenu.js
thomasgwatson: myHome is probably redundant now and likely can be stripped out. It existed before the `my` static group was created.

I feel like we've already been on the 'does this need more abstraction' mery-go-round here and changed it into a presenter and then reverted. So, I'd love to punt on this until later, when things are more settled.
The tree-style ordering happens once in mobile, once in web. And the alphabetical ordering happens... once in mobile and once in web.

lorenjohnson: I certainly don't think we're on a merry-go-round, and don't know what you are referring to in terms of something having been abstracted and reverted. In Mobile there has always been a My static group, that is how it was setup originally. I'm not following entirely, but if you're feeling like just pushing through then please make final amendments and I'll give it another careful review and see if it is all coming together. Do you have any changes you will still make?

thomasgwatson:
Ok, I'm ready to merge this.

Created #245 and #244 to sort out some outstanding issues

packages/hooks/useGetContextWidgetChildren.js
lorenjohnson: Sure, but two caveats: 

1) Please name this hook either `useWidgetChildren` or `useContextWidgetChildren` and return `childWidgets` or simply `widgets`, vs `items`.  I do understand that this will mean touching a few files in Web as well where this is used. In this case I THINK the ContextWidget modeling of a Presenter + a hook (or two) I'm suggesting for Widgets in the UI had already emerged and I'm just wanting to clarify/reinforce this pattern. This should set the stage going forward, and that this is the moment to do that.

2) Read the background below and let me know if it helps you at all getting where I'm coming from in what may be seem picky or unnecessarily blocking around this changeset to date  🙃

---

**Background**

A lot of what may come across as picky pushback from me here at its root arises from me reading the code and feeling unclear about what and why things are happening at various points. Working through the naming of things, iterating it if need be, is how I as a developer come to understand a problem and its solution. I think its also a functional way for us to come into coherency together was a team. 

It seems this new Widget world is going to shape a lot of what is emerging in our apps, and I'm only just now getting a more clear sense of the bones of it. More consistent naming would have helped me get there sooner, and there is still some work to do there I think. 

My most burning questions at the moment, and those which underpin my request for renaming of the hook above:

- Why are we using this `items` nomenclature in the Widget context? I see it a couple places--so perhaps I'm missing what it's signifying? 
- From the perspective of our UI code is there a distinction between a Widget and a ContextWidget? In OO terms does ContextWidget inherit from Widget, or vice versa? Or are they synonyms?

apps/mobile/src/components/ContextMenu/ContextMenu.js
lorenjohnson: Can you please put a todo comment above this line (or wherever you think it best belongs) to the effect of:

"TODO: myHome is probably redundant now and likely can be stripped out. It existed before the my static group was created."

The rest we'll let ride...

lorenjohnson (changes requested):
Please respond to each of the open conversations, with the requested action/change, or reply if we're in disagreement. I think once you digest what I tink are my final remarks on this one, that you'll find the remaining code changes minimal and quick to make.

packages/presenters/ContextWidgetPresenter.js
```diff
@@ -247,60 +254,70 @@ export const COMMON_VIEWS = {
   'ask-and-offer': {
     name: 'Ask & Offer',
     icon: 'Request',
+    iconName: 'Request',
```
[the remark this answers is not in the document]
thomasgwatson: It broke something so Tibet reverted it but I thought then it would break things on mobile, so added both. Looked into it and it was just that the iconNameResolver needed updating

apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
```diff
@@ -7,7 +7,7 @@ import ContextWidgetPresenter from '@hylo/presenters/ContextWidgetPresenter'
 export function WidgetIconResolver ({ widget: providedWidget, style, className }) {
   if (!providedWidget) return null
 
-  const widget = ContextWidgetPresenter(providedWidget, {})
+  const widget = ContextWidgetPresenter(providedWidget, { t: t => t })
```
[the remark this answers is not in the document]
thomasgwatson: Its mocked because it doesn't impact the fields iconResolver depends on.

I don't know if we should `present` here, but instead depend on presented widgets being passed in, which I think pretty consistently happens.

packages/hooks/useGetContextWidgetChildren.js
thomasgwatson: ##### items
items was a more generic name for widgetChildren, childWidgets etc. Because there were cases were the contents of a widget could be dynamically defined,  there was a time when I was thinking all sorts of things might be jammed into the contents of a widget. But now anything that gets passed in as content now gets wrapped in a limited shell of a widget anyway. Its a inform envelope for all sorts of entities. So childWidgets is good to go with now.


##### Widgets vs Context Widgets vs ???
So, in the strictest sense, I would say that the UI component is the widget. The essence of widget, if you will.

When product/design talk about widgets, they are more or less referring to the menu items. The more special/complex/unique/precious a menu item is (the members widget for example), the more widgety it really is.

OO: Contemporary use of widget (anything from the last two years) is just short-hand reference for context widget. There is a prior widgets table, for a view that is basically defunct. Those widgets were *also* an attempt at allowing customization and were heavily connected to the 'farm' type groups. From a product perspective, this feature set basically failed. Those old widgets have been abandoned. Once the dust settles on the redesign (there isn't some surprising emergency community outcry over that feature going AWOL), I have a bookmark to go through and cut out the old widgets code.

apps/web/src/components/WidgetIconResolver/WidgetIconResolver.jsx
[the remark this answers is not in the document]
lorenjohnson: Gotcha. Yes, that is what I thought, and let's just assume the Widget is already presented. If you want a safeguard for now then you could add a `if (providedWidget._presented) throw new Error('Expects a presented Widget') thing at the top. 

Where things are seeming to land with the Group in its Presenter + hook pattern is that we are almost always dealing with a presented entity becuase we mostly only ever get to a group through the hook. I think this is a likely and desirable place to end-up in Widget land too...

packages/hooks/useGetContextWidgetChildren.js
lorenjohnson: Thank you! Very clear description and helpful background. OMG all of the above is SOOOOO helpful to me, more than you know. I am sorry I didn't get to just asking more clear questions sooner. It is a big lesson for me here, to stop and ask more questions first. So much of what was going on here for me was that I didn't see what you saw, and was trying to make sense of it. And pragmatically I feel empowered with these notes to move forward and sort out little things as I come across them and feeling like I'm doing the right things.

PR #222, opened 2025-02-06
Messages area, subscriptions fixes and refinements

lorenjohnson:
Fixes various subtle issues, but not all, for Messages area specifically around incoming new messages and threads reflecting appropriate status in badges. In trying to nail this down I was reminded that there are a variety of small issue with our backend implementation with MessageThreads and Message which keep us from actually being able to make an entirely coherent UX. For now I think this is closer to it, with only a couple small annoyances, which I'm not going to explain here for the sake of time. We can revisit when we have more time, but this is overall an improvement on what is here and worth pushing forward.

### DIFF
--- a/apps/backend/api/graphql/makeSubscriptions.js
+++ b/apps/backend/api/graphql/makeSubscriptions.js
@@ -72,7 +72,9 @@ export default function makeSubscriptions () {
           messageThread.makeModelsType = 'MessageThread'
           return messageThread
         }
-        if (payload?.notification) return new Notification(payload.notification)
+        if (payload?.notification) {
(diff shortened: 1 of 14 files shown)


tibetsprague (approved):
id love to know what you remembered about what needs to change in our backend...

lorenjohnson:
> id love to know what you remembered about what needs to change in our backend...

Indeed. Pretty fried at the moment, but its in me when the times comes. May get to writing something up about it as the dust settles on another day. But basically if we start from the desire UX first I think it will reveal the strangeness and gaps clearly enough.

PR #225, opened 2025-02-07
FlashList Conversion 💥

lorenjohnson:
Conversion was, as hoped and expected, painless and highly beneficial--very needed especially for our Stream. There may/may not be little quirks or edge cases to catch in testing, but generally everything is just better in FlashList land.  Thanks to whomever put this on our radar, it was a good call!

Notes:
- The biggest work here was the re-factor of Comments / ChildComments display which used to be a SectionList
- There are two other SectionLists in the app remaining: Group > Groups, which not usually very intense and can stay that way, and DrawerMenu which is about to be completely re-written... 
- There is some reduction of querying around Comments included in this PR, as well as the re-addition of Post.comments pagination which we evidently lost somewhere in recent re-factors.
- When utilizing for creating new FlashLists the estimatedItemSize is a suggested value that comes from running the component without that configured and basically taking the suggested estimate that FlashList delivers as a warning when you don't give this prop.
- This is a new Pod and the Podfile.lock is up to date of course, but whenever there is a new Podfile.lock update take that as a cue that you need to rebuild the app for things to not break.


apps/mobile/src/components/Comments/Comments.js
```diff
@@ -82,66 +62,64 @@ export const Comments = React.forwardRef(({
...
       // Footer is Header, etc.
       inverted
       ListFooterComponent={Header}
-      renderSectionFooter={SectionFooter}
-      renderItem={Item}
-      sections={sections}
+      renderItem={renderItem}
+      data={comments}
+      // This means that FlashList will re-render anytime this/these values change
+      extraData={highlightedComment}
```
lorenjohnson, line 120: FYI -- if we find any quirks in critical places, likely it will be around other extraData things needing to be added. For now I don't see others. extraData is basically equal to the dependency array in a hook but for items in the list.

apps/mobile/src/components/Comments/Comments.js
```diff
@@ -82,66 +60,64 @@ export const Comments = React.forwardRef(({
     </>
   )
 
-  // Comment rendering (parent)
-  const SectionFooter = ({ section: { comment } }) => {
+  const renderItem = ({ item: comment }) => {
     if (!comment) return null
+    // Inverted flat list so comments will render in reverse order, but child comments
```
tibetsprague, line 65: wait whey are comments rendering in reverse order? we should have older comments at the top and newer ones at the bottom

tibetsprague (approved):

apps/mobile/src/components/Comments/Comments.js
lorenjohnson: One rule in life I have is that there is no way to describe an inverted FlatList without someone getting confused, and usually the one describing it. It is doing the right thing, promise. It's all in the comments in that file. But I will try again to explain here. 

We were using an inverted section list so everything was reversed in the list for comments and child comments, now only the comments are subject to the inversion and each comment renders within its flash list card all its child comments, and this render in non-inverted order. So in this new implementation they initially appeared in the wrong order as we query them in the wrong order for how we want them to display (i.e. "desc", newest up top, older down below). So I reverse it again here vs changing the graphql query which is in the shared directory to query comments desc for the inverted list and their child comments in asc order. Because that would propagate the confusion upstream. So it is just handled here in the component that creates the confusion.

PR #236, opened 2025-02-07
reactOn and deleteReaction mutation updaters

lorenjohnson:
The above named mutations have been troublesome in the URQL setup due to their return type of Post. This PR includes primarily the URQL mutation updaters for both of those cases so they reconcile with the cache without forcing cache invalidation / requery on a paginated Comment set (and its parent post, and usually also the Stream). 

Works well, but would probably be best to update the backend return type to simply return a Reaction which should automatically resolve with the URQL cache without a custom updater. This extra update logic is needed for now though, and works.


thomasgwatson (approved):
lgtm

PR #239, opened 2025-02-08
add search to return url on close

### LINKED ISSUE #238: Return URL missing search params
There are at least two places where we close a dialog or modal: PostDetail and CreateModal (aka PostEditor), both use different methods for navigating back to the previous URL.

----
Here in navigation.js:
https://github.com/Hylozoic/hylo/blob/77ebdecc14d50f0cf286781e67f41e799649f121/apps/web/src/util/navigation.js#L222

Which is called from PostDetail.js:
https://github.com/Hylozoic/hylo/blob/77ebdecc14d50f0cf286781e67f41e799649f121/apps/web/src/routes/PostDetail/PostDetail.js#L133

----
Here in CreateModal:
https://github.com/Hylozoic/hylo/blob/77ebdecc14d50f0cf286781e67f41e799649f121/apps/web/src/components/CreateModal/CreateModal.js#L31

KevinTriplett:
closes #238

Kinda ugly but works to return back to viewmode from whence the user clicks the post (such as calendar but also applies to anything that isn't Card viewmode)

Also restores the sort/filter dropdown selection.

### DIFF
--- a/apps/web/src/components/PostCard/PostHeader/PostHeader.connector.js
+++ b/apps/web/src/components/PostCard/PostHeader/PostHeader.connector.js
@@ -36,7 +36,7 @@ export function mapStateToProps (state, props) {
 
 export function mapDispatchToProps (dispatch, props) {
   const { groupSlug } = props.routeParams
-  const closeUrl = removePostFromUrl(window.location.pathname)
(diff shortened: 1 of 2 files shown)


KevinTriplett:
The other option is to pass `location.href` instead of the concat of `pathname + search`

tibetsprague (approved):

PR #246, opened 2025-02-10
Move GroupWelcomeCheck logic into group presenter; fix group-related …

### LINKED ISSUE #245: Refractor how GroupWelcomeCheck works
See comments on:
https://github.com/Hylozoic/hylo/pull/220#discussion_r1948263251

thomasgwatson:
…issues with GroupWelcomeLanding

Closes #245


packages/presenters/GroupPresenter.js
```diff
@@ -14,10 +14,27 @@ export default function GroupPresenter (group) {
     isPublicContext: group?.slug === PUBLIC_CONTEXT_SLUG,
     isMyContext: group?.slug === MY_CONTEXT_SLUG,
     isAllContext: group?.slug === ALL_GROUPS_CONTEXT_SLUG,
+    shouldWelcome: doesUserNeedToBeWelcomed(group, currentUser),
```
lorenjohnson, line 17: Make `doesUserNeedToBeWelcomed` be `shouldWelcomeResolver`.

packages/hooks/useCurrentGroup.js
```diff
@@ -23,18 +23,19 @@ export function useGroup ({
...
-  const [{ data, fetching, error }, reQuery] = useQuery({
+  const [{ data, fetching: groupFetching, error: groupError }, reQuery] = useQuery({
     ...useQueryArgs,
     query: groupDetailsQueryMaker(groupQueryScope),
     variables: { id: groupId, slug: groupSlug },
     pause
   })
   const rawGroup = contextGroup || data?.group
-  const group = useMemo(() => rawGroup && GroupPresenter(rawGroup, { t }), [rawGroup])
+  const group = useMemo(() => rawGroup && GroupPresenter(rawGroup, { currentUser }), [rawGroup, currentUser])
```
lorenjohnson, line 36: - I think this was a really good call! I want to highlight this case of passing currentUser from the hook to a presenter, which now essentially requires it. This is how they kinda become matched (coupled) sets. A hook ferries the async / UI stuff that the presenter needs to give a properly transformed presenter object and then the hook delivers that enriched object back to the component. This is what I see emerging and am already using in a few other places. I don't like coupling, but where things must be coupled I want the naming to be tightly aligned. This will keep us out of a lot trouble, promise. 

- Please keep passing `t` here even though we're not using it currently in Presenter, and if you don't agree with that then remove useTranslation from this hook as I don't think it's used otherwise. 

_Pretty sure I will push forward a PR which makes i18n a shared module (just the instantiation) between both platforms soon, and then we won't need to pass it at all as we'll just grab it in the Presenters..._

packages/presenters/GroupPresenter.js
lorenjohnson, line 17: Make `doesUserNeedToBeWelcomed` be `shouldWelcomeResolver`.

apps/mobile/src/components/GroupWelcomeCheck/GroupWelcomeCheck.js
```diff
@@ -5,25 +5,15 @@ import useCurrentGroup from '@hylo/hooks/useCurrentGroup'
 
 export default function GroupWelcomeCheck () {
   const navigation = useNavigation()
-  const [{ currentUser }] = useCurrentUser()
```
lorenjohnson, line 8: Remove useCurrentUser import, and please get your linting working :) This file still imports useCurrentUser though it doesn't utilize it anymore.

lorenjohnson (changes requested):
I think this logic should go higher up the nav stack in the ContextMenu where we otherwise handle all things context and context switching related:

```
export default function ContextMenu () {
  const { myHome, groupSlug } = useRouteParams()
  const [{ currentGroup, fetching }] = useCurrentGroup({ setToGroupSlug: groupSlug })
  const currentGroupSlug = useCurrentGroupSlug()
  const contextWidgets = currentGroup?.contextWidgets || []
  const orderedWidgets = useMemo(() =>
    orderContextWidgetsForContextMenu(contextWidgets.map(widget => ContextWidgetPresenter(widget, { t }))), [contextWidgets, currentGroup, myHome]
  )

  useEffect(() => {
    if (!fetching) {
      if ((currentGroup.shouldWelcome)) {
        navigation.navigate('Group Welcome', { groupId: currentGroup?.id })
      }
    }
  }, [currentGroup, fetching])
...
```

If for some reasons that doesn't work, then let me know what you find. Otherwise, let's do this, remove the GroupWelcomeCheck component--and please do lint.

Issue #254, opened 2025-02-12
`my` context on mobile: User settings and my/all content views


thomasgwatson:
- [x] Allow `my` context to be accessed from the user avatar
- [x] Make sure that all the `my` and `all` views are linked and available from the `my` context menu
- [x] Remove user settings upper tab bar

PR #256, opened 2025-02-12
'my' context and user settings in mobile

thomasgwatson:
See #254 for the goals of this branch.

@lorenjohnson I'm trying to treat each commit on this branch as a diff worth discussing.

First one in, want to get your feel on the brute force approach for my context nav out of the tab

Will next focus on making sure the various `my` context menu links work


thomasgwatson:
Ok, one piece of possible weirdness:

In the front end, we don't have the static groups. And the front-end basically treats `my` and `all` as the same entity. So similarly.... we should probably collapse those two down to the same static group?

apps/mobile/src/navigation/TabsNavigator.js
```diff
@@ -77,7 +86,14 @@ export default function TabsNavigator () {
       />
       <Tabs.Screen
         name='Settings Tab'
-        component={UserSettingsTabsNavigator}
+        component={HomeNavigator} // it will never navigate to this but we need to pass a valid component here anyway
```
lorenjohnson, line 89: This looks fine by me for now.

lorenjohnson:
> Ok, one piece of possible weirdness:
> 
> In the front end, we don't have the static groups. And the front-end basically treats `my` and `all` as the same entity. So similarly.... we should probably collapse those two down to the same static group?

By front-end you mean Web. How does it relate to things here? I am not making the connection, but I do hear that My and All are the same, or mostly the same target. I think if they are not actually the same then keeping them as their own targets it still good, and to share their common functionality through components they both use, vs blending the logic. However I'm speaking on abstractly and not entirely familiar with the patterns yet. My includes all of All though I think you were telling me? But still have an /all route?

thomasgwatson:
So yeah...

Post new context menu; `my` has basically subsumed all of `all` routes... so now I'm wondering if we don't just go the whole way and change all references to `all` to `my`.

On web, at the moment, while the context part of their paths are still different, both the `my` and `all` paths end up with the same context menu. And obviously display different things in the main panel based on the rest of their paths.

Its probably more trouble to keep both `all` and `my` than it will be to tidy up the remainder of `all` references. 

@tibetsprague would like you to add your perspective; do we really need to have the `/all` context?

apps/mobile/src/navigation/UserSettingsTabsNavigator.js
```diff
@@ -72,7 +72,7 @@ export default function UserSettingsTabsNavigator ({ navigation, route }) {
     })
   }, [navigation, route])
 
-
+  console.log('does this get called? at alll?')
```
thomasgwatson, line 75: It doesn't, because the navigator isn't rendered

apps/mobile/src/navigation/UserSettingsTabsNavigator.js
```diff
@@ -85,42 +85,42 @@ export default function UserSettingsTabsNavigator ({ navigation, route }) {
         name='Edit Profile'
         component={UserSettingsWebView}
         initialParams={{
-          path: '/settings'
+          path: '/my/edit-profile'
```
thomasgwatson, line 88: I was trying to navigate to these webviews with the existing structure and haven't figured it out. But do they even need a separate settings tab/navigator now? Could these screens just be moved over to the home navigator?

apps/mobile/src/navigation/linking/index.js
```diff
@@ -36,16 +36,27 @@ export const routingConfig = {
...
 
-  // context group routes (/all, /public, /my)
-  '/:groupSlug(all|public)':                                 `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
+  // context group routes (/all, /public)
   '/:groupSlug(all|public)/post/:id':                        `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Post Details`,
   '/:groupSlug(all|public)/post/:id/comments/:commentId':    `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Post Details`,
   '/:groupSlug(all)/members/:id':                            `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Member`,
   '/:groupSlug(all)/topics/:topicName':                      `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
+  '/:groupSlug(all|public)/events':                          `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
+  '/:groupSlug(all|public)':                                 `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
```
thomasgwatson, line 45: Moved this to the bottom, so other all/public paths wouldn't be eaten by this one first

apps/mobile/src/navigation/linking/index.js
```diff
@@ -36,16 +36,27 @@ export const routingConfig = {
...
   '/:groupSlug(my)/mentions':                                `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Mentions`,
   '/:groupSlug(my)/announcements':                           `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Announcements`,
+  // my settings
+  '/:groupSlug(my)/edit-profile':                            `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Settings Tab/Edit Profile`,
+  '/:groupSlug(my)/account':                                 `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Settings Tab/Account`,
+  '/:groupSlug(my)/notifications':                           `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Settings Tab/Notifications`,
+  '/:groupSlug(my)/blocked-users':                           `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Settings Tab/Blocked Users`,
+  '/:groupSlug(my)/:section?':                               `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Settings Tab/Edit Profile`,
+  // So, some user settings aren't explicitly linked, but I assume are avaiable because of this section wildcard.
+  // But ultimately everything just redirects to the `Edit Profile` screen.... so maybe none of the extra paths need to be here?
```
thomasgwatson, line 59: Right?

apps/mobile/src/screens/UserSettingsWebView/UserSettingsWebView.js
```diff
@@ -7,14 +7,15 @@ import useCurrentUser from '@hylo/hooks/useCurrentUser'
 
 export default function UserSettingsWebView ({ path: pathProp, route }) {
   // TODO: URQL! - Untested, intention is to refresh cache
+  console.log('UserSettingsWebViewwwsssswww', pathProp, route) // This never gets called when navigating from the context menu
```
thomasgwatson, line 10: Couldn't seem to navigate to this

thomasgwatson:
@lorenjohnson 

This is what openURL is translating the path to... and it looks right, as far as I can tell. But when its passed to `return navigation.dispatch(actionForPath)`, it doesn't seem to do anything. 

From the openURL debugging:
Screenshot 2025-02-13 at 3 32 08 PM

Got any ideas on what needs to be switched around here? Am I just barking up the wrong tree by trying to navigate to these screens after changing the path structure?

thomasgwatson:
oooooooooo I have an idea....

I removed the Settings Tab from the Tabs navigator when I made the changes to the user avatar... will fiddle with that for a second

thomasgwatson:
Ok, that kinda works. I'll keep playing with this when I get back from my break

tibetsprague:
my perspective on /my and /all. The main reasons to keep /all are: 
1. To support old links that use /all. 
2. Because it becomes awkward to figure out the right URL to all posts across all your groups when it starts with /my. Would it be /my/all-group-posts? /my/all-group-events, etc? /all/posts is just cleaner and makes sense when you read it. 

That being said I could be convinced to switch to /my for everything with a good proposal :)

thomasgwatson:
Yeah, I'm back at the /my/ and /all/ mixed context situ. Mobile relies on modeling those contexts as a Group interface (same data structure as a group more or less). Currently, the profile/user menu (`my` context menu) allows access to both /all and /my routes.... that works fine with web, because it just responds to the change in the url. But if I switch between different routes on mobile, that end up on the Stream screen, it doesn't necessarily know to refresh the stream because that switch in route doesn't change the static context from (for example) /my to /all .

So, maybe I am less about trashing /all routes, and now wondering @lorenjohnson, does it make sense to have some part of openURL that is looking at the :context part of the path and calling changeToGroup. That will make sure that any paths being handed into the app is forced to sync the right :context

What do you think?

tibetsprague:
ill note that if it makes things easier I am ok with /my/groups/stream, /my/groups/map, etc instead of /all

lorenjohnson:
> Yeah, I'm back at the /my/ and /all/ mixed context situ. Mobile relies on modeling those contexts as a Group interface (same data structure as a group more or less). Currently, the profile/user menu (`my` context menu) allows access to both /all and /my routes.... that works fine with web, because it just responds to the change in the url. But if I switch between different routes on mobile, that end up on the Stream screen, it doesn't necessarily know to refresh the stream because that switch in route doesn't change the static context from (for example) /my to /all .
> 
> So, maybe I am less about trashing /all routes, and now wondering @lorenjohnson, does it make sense to have some part of openURL that is looking at the :context part of the path and calling changeToGroup. That will make sure that any paths being handed into the app is forced to sync the right :context
> 
> What do you think?

I'll have to re-read, but pretty sure on first pass this is all stuff that should be handled at the linking table level... Not fully following though so will re-read a couple more times and loop back if something becomes more clear. In the mean time any more concrete details with some specific paths would help.

lorenjohnson:
Are there any /all routes that don't have My as their context menu?

tibetsprague:
No not any more

lorenjohnson:
> ill note that if it makes things easier I am ok with /my/groups/stream, /my/groups/map, etc instead of /all

we can definitely work around what is here on mobile without a big fuss, but by I think those paths are a lot more clear and think we should go with them. not a biggy making the switch on the Web side?

thomasgwatson:
current /my and /all changes:
- basically I got around this from a query/data perspective by pulling the `:context` out of the `useRouteParams`, instead of relying on `currentGroup.slug`

I think there might be sense in merging /my and /all more fully but I'd defer messing with that before the redesign

packages/hooks/useStreamQueryVariables.js
```diff
@@ -24,11 +25,11 @@ export default function useStreamQueryVariables ({
       : null,
     childPostInclusion: currentUser?.settings?.streamChildPosts || 'yes',
     collectionToFilterOut: null,
-    context: isContextGroupSlug(forGroup?.slug)
-      ? forGroup.slug
+    context: context || (isContextGroupSlug(forGroup?.slug) // Relying on context now but will fall back to prior logic
```
lorenjohnson, line 28: This is likely to trip-up the currentGroup state, and I need to chat about this to understand if this is the only and best option and why.

apps/mobile/src/navigation/linking/index.js
lorenjohnson: Correct! Always be on the lookout for that, the logic can seem a little inverted if thinking it through. This is always the correct pattern though. The least specific things last.

packages/presenters/GroupPresenter.js
```diff
@@ -31,23 +31,30 @@ function shouldWelcomeResolver (group, currentUser) {
...
       isAllContext: group.slug === ALL_GROUPS_CONTEXT_SLUG,
-      profileUrl: ''
+      profileUrl: profileUrlResolver(currentUser)
     })
   }
   // TODO redesign: Presented widgets, but they need a t object (or to add global--see note in ContextWidgetPresenter)
   return (group?.contextWidgets?.items || []).map(ContextWidgetPresenter)
 }
 
+// Until such time as we have navigation helpers in a shared context,
```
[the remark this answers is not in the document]
lorenjohnson: To reduce confusion and coupling here, I think this helper should either be created and used in the one place its currently used, presumably ContextMenu, and the currentUserProfilePath or simply profilePath is the thing to pass to this method from the component. If it is used more generally then we put it in util/navigation.js, and as a "url" helper we pass it either the whole user or just the id depending on whatever is standard practice there. Maybe this is the merry-go-round you were referring to, but I'm wondering if contextWidgetsResolver logic combined with the ordering logic which I think can/should be happening at that same place, is all put in a hook which returns what ContextMenu needs. The convergence zone. The open question for me is whether this transformation and data prep, childWidgets included, has a use case outside of the menu... Because if it does that probably changes my opinions on this.

packages/presenters/GroupPresenter.js
```diff
@@ -31,23 +31,30 @@ function shouldWelcomeResolver (group, currentUser) {
...
   const agreementsChanged = (!isContextGroupSlug(group?.slug) && numAgreements > 0) &&
     (!agreementsAcceptedAt || agreementsAcceptedAt < group?.settings?.agreementsLastUpdatedAt)
-  
+
   return ((!isContextGroupSlug(group?.slug) && showJoinForm) || agreementsChanged || (group?.settings?.askJoinQuestions && !joinQuestionsAnsweredAt))
- }
+}
 
-function contextWidgetsResolver (group) {
+function contextWidgetsResolver (group, { currentUser }) {
```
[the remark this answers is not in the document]
thomasgwatson: So, there are basically two different sortings of context widgets for display.

Do we... have two copies of the widgets on each group, one each for the different contexts? Tibet is now basically writing the `isHiddenInAllView` counterpart in another PR.

We can't do all of our presenting as soon as the Group is loaded; the contextual (regular english, no symbol) placing of the group/widget matters, which we only know downstream. We can't double present after we have the placing, as it will just return the first, limited presentation. We could add more state to each presented widget, say widget._placed and check for that as well. 

I don't think its a bad thing, that when you get a collection, that you then shape it for the needs of its place-specific usage.

packages/presenters/GroupPresenter.js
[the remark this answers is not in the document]
thomasgwatson: Currently there are only two places context widgets for a group are shown: the context menu and the all view. Roughly speaking, we receive an unsorted flat array of context widgets from the server. And we want a numerically ordered tree and an alphabetically ordered flat array for the two different contexts.

Ok, jumping back from https://github.com/Hylozoic/hylo/pull/256#discussion_r1960643566
I said "Do we... have two copies of the widgets on each group, one each for the different contexts?" because that discussion was going on in GroupPresenter.

But made what I read here is that no, we create a hook... this hook requires all of the contextual things like currentUser, t, and anything else that is needed. And it exposes two sets of transformed and presented context widgets, just does it all in the one go. And then the consumer of this hook can just choose which set of widgets it needs for its context. We'd just need to not pre-present widgets before this point (like we currently do in GroupPresenter, which doesn't currently have access to `t`, which means i18n isn't working for widget titles)

What about that?

packages/presenters/ContextWidgetPresenter.js
```diff
@@ -134,7 +126,8 @@ function widgetTypeResolver ({ widget }) {
   )
 }
 
-const isHiddenInContextMenuResolver = (widget) => {
+export const isHiddenInContextMenuResolver = (widget) => {
+  // This doesnt work properly if evoked before the widgets are ordered.
```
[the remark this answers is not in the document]
thomasgwatson: One of the AND rules for hiding (something needs to fail a set of the rules to be hidden), is not having any children. Since widgets arrive from the server in a flat array, the needs-to-be-hidden quality of a widget has to wait until things are ordered for this to be accurate.

Why not expose this as a function on widget? I tried exposing it as a function and calling it in the two or so places its used across mobile and web. Even when I was guarding against unpresented widgets, it was still hitting a ' is not a function'. It was just simpler to import the function and pass the widget into.

apps/mobile/src/screens/GroupSettingsWebView/GroupSettingsWebView.js
```diff
@@ -1,23 +1,27 @@
...
+import { alabaster, amaranth, capeCod, rhino40, rhino80 } from 'style/colors'
 
-export default function GroupSettingsWebView () {
+export default function GroupSettingsWebView ({ path: pathProp, route}) {
   const webViewRef = useRef()
+  const [, queryCurrentUser] = useCurrentUser({ requestPolicy: 'network-only', pause: true })
   const { groupSlug, originalLinkingPath, settingsArea: routeSettingsArea } = useRouteParams()
   const [selectedSettingsArea, setSelectedSettingsArea] = useState(routeSettingsArea)
-
+console.log('!! routeSettingsArea', routeSettingsArea, pathProp, route)
```
thomasgwatson, line 13: @lorenjohnson do you still want this?

packages/presenters/GroupPresenter.js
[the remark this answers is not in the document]
lorenjohnson: - The hook I'll put up in PR shortly is going to be useContextMenuData for ContextMenu, it is the Presenter for the ContextWidget data, if you will. I will write it somewhat like useCurrentGroup where there are compositional hooks which can then also be used for useAllViewsData. I can either put that hook next to the AllViews screen, or just keep it in the same hook file and call that module something general about Widgets. So this is similar to what you're saying, but I will be trying to not "put it all in one hook", as the goal here is to create a clear place to define, present/transform and otherwise shape whatever we need to make a more simply UI ContextMenu component. And yes, we can do the same for AllViews and I suspect you're suggesting what you are because the data prep needed for both most have a lot in common... I will still be writing the hook in such a way as to not render anything, leaving any rendering and such to the components.
- Purely in terms of the UI what is a Widget vs a ContextWidget? I know I've asked before but this still isn't clear. I'd like to use ContextWidget and Widget synonymously with a slight preference for simply Widget if the "Context" part of ContextWidget is not at this point qualifying anything. If it is I am, in order to make clear stuff, needing to still understand the difference--even if it is architecturally leaving space for something to come later.
- [This PR](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) fixes the i18n access by making its instance a shared module, and then we can in all places access i18n directly either through the useTranslation hook in a component or hook, or through import i18n from @hylo/shared/i18n. We then can eliminate the cross dependency of the passing of t and such, which really shouldn't be needed.

packages/presenters/GroupPresenter.js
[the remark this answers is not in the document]
lorenjohnson: I'm losing the thread on this one, but the hook I'll forward will probably subsume this level of data prep and take it off both the Group and Widget Presenters which I think will get thinner. Would like to know if the`isHiddenInAllVew` thing Tibet is working relates here, and if it is being constructed as something we share? Until I get more clear about that I won't worry about it for now, and just work with what I'm seeing in the Mobile code...

apps/mobile/src/screens/GroupSettingsWebView/GroupSettingsWebView.js
lorenjohnson: Nope

packages/presenters/GroupPresenter.js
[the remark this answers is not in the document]
thomasgwatson: Tibet has added widget hiding logic for the AllView that is parallel but different to the widget hiding logic for the context menu. It is not in a hook or anything at the moment, just in the component; and I'm not sure if its on dev yet.

And yes, the dedicated hook concept for this will cover the use case

PR #257, opened 2025-02-12
Mobile test setup fixes and linking tests

lorenjohnson:
- Catchs-up Mobile test setup to not be completely broken including a whole lotta jest and babel juggling
- Adds `getStateToPath.test.js` which iterates actual linking paths and generates the screen path and params they route to. This is a first iteration but does most of everything we could hope/need to nail our linking routing and be alerted if something goes wrong (once we're running CI again in particular).


package.json
```diff
@@ -36,6 +36,8 @@
     "@luma.gl/constants": "9.0.28",
     "@luma.gl/gltf": "9.0.28",
     "@emotion/utils": "^1.1.0",
-    "@emotion/sheet": "^1.1.0"
+    "@emotion/sheet": "^1.1.0",
+    "htmlparser2": "9.1.0",
```
lorenjohnson, line 40: NOTE: This a version force for a couple of libraries used in @hylo/shared for htmlToText. The dependency is shared was lower than this, and not necessarily compatible at least in smear terms (I think it requires 7.x of htmlparser2). Nothing seems to be broken, and after to way too much time into I cam to the conclusion that there was no better way.

apps/mobile/src/util/testing/index.js
lorenjohnson, line 1: This updated MSW setup is largely untested, but not crashing. Looking forward to trying it out soon.

apps/mobile/src/navigation/linking/index.js
```diff
@@ -82,23 +83,23 @@ export const routingConfig = {
...
   // catch-alls
-  '(.*)/group/:groupSlug':                                   `${AUTH_ROOT_SCREEN_NAME}/${modalScreenName('Group Explore')}`,
-  '(.*)/group/:groupSlug/explore':                           `${AUTH_ROOT_SCREEN_NAME}/${modalScreenName('Group Explore')}`,
-  '(.*)/members/:id':                                        `${AUTH_ROOT_SCREEN_NAME}/${modalScreenName('Member')}`,
-  '(.*)/post/:id':                                           `${AUTH_ROOT_SCREEN_NAME}/${modalScreenName('Post Details')}`,
-  '(.*)/post/:id/comments/:commentId':                       `${AUTH_ROOT_SCREEN_NAME}/${modalScreenName('Post Details')}`,
-  '(.*)/create/group':                                       `${AUTH_ROOT_SCREEN_NAME}/Create Group`,
-  '(.*)/create/post':                                        `${AUTH_ROOT_SCREEN_NAME}/Edit Post`,
-  '(.*)/post/:id/edit':                                      `${AUTH_ROOT_SCREEN_NAME}/Edit Post`,
+  ':unmatchedBasePath(.*)/group/:groupSlug':                  `${AUTH_ROOT_SCREEN_NAME}/${modalScreenName('Group Explore')}`,
```
lorenjohnson, line 86: now names the unmatched part of the path and also sends that along to screen params.

apps/mobile/config/jest/setupTests.js
```diff
@@ -3,6 +3,7 @@ import 'react-native-gesture-handler/jestSetup'
 import mockSafeAreaContext from 'react-native-safe-area-context/jest/mock'
 import mockClipboard from '@react-native-clipboard/clipboard/jest/clipboard-mock.js'
 
+// console.log(require.resolve('urql'));
```
thomasgwatson, line 6: ??

apps/mobile/src/navigation/linking/__snapshots__/getStateFromPath.test.js.snap
```diff
@@ -0,0 +1,184 @@
...
+  "/groups/rogue-scholars/chat/philosophy/post/12345": {
+    "path": "AuthRoot/Post Details - Modal",
+    "routeParams": {
+      "id": "12345",
+      "originalLinkingPath": "/groups/rogue-scholars/chat/philosophy/post/12345",
+      "pathMatcher": ":unmatchedBasePath(.*)/post/:id",
+      "unmatchedBasePath": "/groups/rogue-scholars/chat/philosophy",
+    },
+  },
+  "/groups/rogue-scholars/chat/philosophy/post/12345?commentId=5555": {
```
thomasgwatson, line 23: should it be `chats`, not `chat` here?

I have some recollection of fixing that on the navigation url helper... on some branch... but maybe it never made it to dev?

apps/mobile/src/navigation/linking/getStateFromPath.test.js
```diff
@@ -0,0 +1,150 @@
+import getStateFromPath from 'navigation/linking/getStateFromPath'
+
+// // Transcription of "notifications paths" from Web
+// '/my/account',
```
thomasgwatson, line 4: Was this the only /my/ path you found or just testing only this one?

apps/mobile/src/navigation/linking/index.js
thomasgwatson: oh, so we can feed that into url sharing, error reporting and analytics?

thomasgwatson (approved):
lgtm

apps/mobile/src/navigation/linking/__snapshots__/getStateFromPath.test.js.snap
lorenjohnson: Oh yes, the current test cases in there are just a transcription form the spreadsheet, plus some. This should be the place we put any and all routes we want to check and have as our standard tests. If I get to it I'll make a little video about how to utilize this test as it has a couple tricks up its sleeve and more if we need them...

apps/mobile/src/navigation/linking/getStateFromPath.test.js
lorenjohnson: Yeah, again the actual test cases in here are, other than the ones transcribe from the spreadsheet relatively arbitrary. Add things or take things out :) If Tibet gets us a master list we can just put them all in here.

PR #266, opened 2025-02-14
Mobile Messages fixes and improvements

lorenjohnson:
- Makes a variety of needed refinements to the Messages UX, entirely in this case within New Message and a Message Thread with no changes to the Thread Index view.
- Makes the Message Input allow multiline with a reasonably tall max height and scrollable after that. Some notable styling tweaks as well. It's looking and working well.
- Makes participant selection inline instead of via a modal, by splitting ItemSelectorModal into a two components so that ItemSelector could be used directly. Will be potentially useful for other things like upcoming Post Editor changes.
- Adds confirmation on close (back button) to New Message such that if you've selected any participants or have a message written you have to confirm that you are ready to discard the message in progress
- Fixes a couple subtle issues in ItemSelector such that default items sometimes didn't show up.
- Adds some minimal way into styling the ItemSelector to accommodate its use in New Message.
- Simplifies and updates the look of Messages (for now), but includes making the Tab Bar background white (was rhino10), as well as the headers in the Messages area. Also employs some Lucide icons in these areas.

Post game notes:

- Most everything looks and feel a whole lot better in Mobile Messages. And ya to some finally visible work!
- Not tested in Android yet (will as part of overall testing, or sooner)
- There are still a couple remaining in the Messages UX to get around to, but nothing stands out to me at this point except how the update subscription for a New Message or Message Thread makes the FlashList flash a re-render when they come in. This is something I'll want to resolve in particular for the New Message update as the primary use case of this stream is so you can be chatting live with someone in a thread, and that doesn't feel great if there are a lot of flashing re-renders...


apps/mobile/ios/Gemfile
```diff
@@ -5,6 +5,6 @@ source 'https://rubygems.org'
 ruby '2.7.5'
 
 # Exclude problematic versions of cocoapods and activesupport that causes build failures.
-gem 'cocoapods', '>= 1.13', '!= 1.15.0', '!= 1.15.1'
+gem 'cocoapods', '>= 1.15.2'
```
lorenjohnson, line 8: Fixes the cocoapods version tug-of-war, for now.

apps/mobile/src/components/CommentEditor/CommentEditor.js
```diff
@@ -1,21 +1,21 @@
...
 import { AnalyticsEvents, TextHelpers } from '@hylo/shared'
-import mixpanel from 'services/mixpanel'
-import { isIOS } from 'util/platform'
 import createCommentMutation from '@hylo/graphql/mutations/createCommentMutation'
+import { firstName } from '@hylo/presenters/PersonPresenter'
 import HyloEditorWebView from 'components/HyloEditorWebView'
-import styles from './CommentEditor.styles'
 import Icon from 'components/Icon'
-import { firstName } from '@hylo/presenters/PersonPresenter'
+import styles from './CommentEditor.styles'
```
lorenjohnson, line 19: Just doing some import clean-up here

tibetsprague (approved):
Whoo! good stuff

thomasgwatson (approved):
lgtm

If you wanted to take it a step further, you could switch the coloring over to nativewind (via className), based on how web messages are colored.

lorenjohnson:
> lgtm
> 
> If you wanted to take it a step further, you could switch the coloring over to nativewind (via className), based on how web messages are colored.

Yes for sure. I specifically am holding off on that for the moment so we don't end-up with a currently somewhat coherent Mobile app design + a partially implemented new design. That said, the Messages UI updates bring it closer as it is and there was some consideration of the bones of things in the updates. That in combination with the functional and code stuff that is cleared-up in this area will also make that process go more quickly. 

_So very much on my mind too, though I would continue to strongly encourage us to not let Mobile theming creep into the scope for  a ~March 8th release because I don't think we can do it well in this time frame, and could shortly after when the dust settles as it is the sort of project that could be most fun and benefit from a more singular focus. Needless to say that doesn't if you are finding time for it before then to not start making progress, but I'd want whatever progress we make to be really targeted and completer (and maybe the Messages area would be a good candidate for that) or otherwise left in its own branch until it was._

PR #267, opened 2025-02-15
Multiday calendar events

KevinTriplett:
Just repeat the event once per day -- sooo may not be obvious that it's a multi-day event, looks more like a recurring event.


KevinTriplett:
This PR is replaced by #288

PR #268, opened 2025-02-15
Updates OneSignal setup

lorenjohnson:
Based on #266 (but probably wouldn't have to be)

Fixes, hopefully, any OneSignal issues, but still not fully confirmed. It is all in the right direction either way so would like to merge it down.


tibetsprague (approved):

thomasgwatson (approved):
lgtm

PR #272, opened 2025-02-16
Improvements to post routing

tibetsprague:
Ensure notifications for chat posts go to the chat room, and others go directly to the post.

Clean up how we link to posts on the front-end creating new useViewPostDetails hook.

Switch from adding comments/:commentId to jump to a specific omment to ?commentId=XXX

### DIFF
--- a/apps/web/src/routes/MapExplorer/MapExplorer.js
+++ b/apps/web/src/routes/MapExplorer/MapExplorer.js
@@ -343,11 +343,11 @@ function MapExplorer (props) {
     dispatch(saveSearch(attributes))
   }, [context, currentBoundingBox, currentUser?.id, dispatch, filters, groupSlug])
 
-  const showDetails = useCallback((postId) => dispatch(navigate(postUrl(postId, { ...routeParams, view: 'map' }, getQuerystringParam(['hideDrawer', 't', 'group'], location)))), [dispatch, navigate, routeParams, location])
+  const showDetails = useCallback((postId) => navigate(postUrl(postId, { ...routeParams, view: 'map' }, getQuerystringParam(['hideDrawer', 't', 'group'], location))), [navigate, routeParams, location])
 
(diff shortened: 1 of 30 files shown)


apps/web/src/routes/MapExplorer/MapExplorer.js
```diff
@@ -695,8 +695,6 @@ function MapExplorer (props) {
   const { hideNavLayout } = layoutFlags
   const withoutNav = isWebView() || hideNavLayout
 
-  const locationParams = location !== undefined ? getQuerystringParam(['zoom', 'center', 'lat', 'lng'], location) : null
```
thomasgwatson, line 698: Where did these go?

apps/web/src/util/navigation.js
```diff
@@ -195,7 +216,7 @@ export function widgetUrl ({ widget, rootPath, groupSlug: providedSlug, context
   } else if (widget.viewUser) {
     url = personUrl(widget.viewUser.id, groupSlug)
   } else if (widget.viewPost) {
-    url = postUrl(widget.viewPost.id, { groupSlug, context })
+    url = postUrl(widget.viewPost, { groupSlug, context })
```
thomasgwatson, line 219: Did we double check that the ContextWidget queries on type Group are returning everything needed for the new postUrl, now that it depends on more than id? The query might be partial

thomasgwatson (approved):
lgtm, just curious about the locationParams on the map views; from recollection, some of that functionality wasn't working and I wrote some tickets to follow up on that. So not expecting you to address the bug here but curious why the locationParams were removed here

apps/web/src/routes/MapExplorer/MapExplorer.js
tibetsprague: the new useViewHeaderDetails hook just pulls in all existing query params into it. thats all this was being used for

apps/web/src/util/navigation.js
tibetsprague: oops, this was supposed to stay as id. postUrl didnt actually end up changing. i messed with it at first and then just created the new primaryPostUrl function

PR #275, opened 2025-02-17
All context gone, switched up to my

tibetsprague:
Did a first pass and generally seems to be working. But it feels like there's too much to test here for me to confidently say lets go for it. Unless we really need this for mobile to work smoothly i say we hold off on it.

### DIFF
--- a/apps/backend/api/services/Frontend.js
+++ b/apps/backend/api/services/Frontend.js
@@ -113,7 +113,7 @@ module.exports = {
     },
 
     mapPost: function (post, context, slug) {
-      let contextUrl = '/all'
+      let contextUrl = '/my/groups'
 
       if (context === 'public') {
         contextUrl = '/public'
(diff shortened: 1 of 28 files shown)


lorenjohnson:
@tibetsprague in working with Tom on the Mobile routing side of this I have just internally in Mobile re-mapped all "all" context to "my" context. This resulted in us having these two routes in Mobile which are internal, `/my/stream` and `/my/map` (were /all/stream and /all/map). The more I look at that the more I like it. Unless we're holding out for something else for `/my/stream` I think this is perfect and now I like it better than `my/groups/*` as I thin that my groups is just a fine thing to imply for `/my/stream`... So just mentioning here in case that change would be any simpler to make... I think Tom is a fan as well of this version. @thomasgwatson you can cat your vote here if you wish?

tibetsprague:
hmm. I mean right now we have /my/posts listing all the posts you created, but that could just as easily be /my/stream, and I was considering whether to make that change. I think it's confusing to have /my/events mean the events from all my groups when it could just as easily mean the events I created, or the events I have marked myself as attending. So im not actually on board with that and I think we should stick with /all for now

thomasgwatson:
I think we did this, right? Can we close this PR?

tibetsprague:
no we have not done this. i merged in the latest dev, see all the changes. this is still a thing

PR #280, opened 2025-02-18
Add Stream, discussion and resources widgets automatically

tibetsprague:
@thomasgwatson see the TODO. is there a reason why we arent just checking whether a widget has an order to see whether to auto add it or not? Because if a widget has been manually added then we probably dont want to auto add it right? Also then we could probably get rid of the auto add column completely

### DIFF
--- a/apps/backend/api/models/Group.js
+++ b/apps/backend/api/models/Group.js
@@ -289,14 +289,8 @@ module.exports = bookshelf.Model.extend(merge({
       .query({ where: { 'posts.active': true } })
   },
 
-  postCount: function () {
-    return Post.query(q => {
-      q.select(bookshelf.knex.raw('count(*)'))
-      q.join('groups_posts', 'posts.id', 'groups_posts.post_id')
-      q.where({ 'groups_posts.group_id': this.id, active: true })
-    })
-      .fetch()
-      .then(result => result.get('count'))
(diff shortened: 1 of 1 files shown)


apps/backend/api/models/Group.js
```diff
@@ -949,9 +943,38 @@ module.exports = bookshelf.Model.extend(merge({
...
+                addToEnd: true,
+                trx
+              })
+            }
+          }
+
           // Check events
           if (postType === 'event') {
             const eventsWidget = widgets.find(w => w.get('view') === 'events')
+            // TODO: instead of checking auto_added shouldnt we just check order? to see if it is added anywhere already?
```
thomasgwatson, line 977: If a steward removes one of these sorts of views from the menu, it will have no order. So this check will put it back into the menu, overriding the stewards update. Basically any mutation on a widget will update its auto_added value,  under the assumption that once a steward has moved or changed it, that they have a specific desire for it to be or not be somewhere, and thus we shouldn't reorder it.

thomasgwatson (approved):
lgtm, noted your TODO and I think no change is desired. Please let me know if that explanation is sufficient or if you think there is a good place for us to document menu behaviour in relation to auto_added

tibetsprague:
cool yeah makes sense. we could also check order but ill leave it as is for now

PR #281, opened 2025-02-18
Only display widgets with views in the All View, except when in edit …

tibetsprague:
…menu mode

Also when editing dont display widgets that have already been added to the menu

And when in edit mode change the name of the All View to indicate edit mode

### DIFF
--- a/apps/web/src/routes/AllView/AllView.jsx
+++ b/apps/web/src/routes/AllView/AllView.jsx
@@ -92,7 +92,7 @@ export default function AllViews () {
   const visibleWidgets = useMemo(() => {
     return contextWidgets.filter(widget => {
       // When not editing only show widgets with a related view or chat room
-      if (!isEditing && !widget.view && widget.type !== 'chat') return false
+      if (!isEditing && !widget.view && !widget.customView && widget.type !== 'chat') return false
       // When editing only show widgets that have not already been added
       if (isEditing && widget.order) return false
       if (widget.visibility === 'admin' && !canAdminister) return false
--- a/apps/web/src/routes/AllView/AllView.jsx
+++ b/apps/web/src/routes/AllView/AllView.jsx
@@ -56,15 +56,16 @@ export default function AllViews () {
   const dispatch = useDispatch()
   const routeParams = useParams()
 
-  // Access the current group and its contextWidgets
-  const group = useSelector(state => getGroupForSlug(state, routeParams.groupSlug))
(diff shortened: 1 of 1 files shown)


apps/web/src/routes/AllView/AllView.jsx
```diff
@@ -94,15 +88,23 @@ export default function AllViews () {
...
-  const widgetsSorted = useMemo(() => {
-    return visibleWidgets.map(widget => {
-      return ContextWidgetPresenter(widget, { t })
-    }).sort((a, b) => a.title.localeCompare(b.title))
-  }, [visibleWidgets])
+  // Filter and sort widgets and get them ready for display
+  const visibleWidgets = useMemo(() => {
+    return contextWidgets.filter(widget => {
+      // When not editing only show widgets with a related view or chat room
+      if (!isEditing && !widget.view && widget.type !== 'chat') return false
```
thomasgwatson, line 95: Just checking this is desired:
- Custom views that have been removed from the menu will be hidden from regular users in all way

apps/web/src/routes/AllView/AllView.jsx
```diff
@@ -94,15 +88,23 @@ export default function AllViews () {
...
+      // When editing only show widgets that have not already been added
+      if (isEditing && widget.order) return false
```
thomasgwatson, line 97: One trap that exists, something I have been meaning to bring up;

You can remove a widget that has children from the menu, and those children are still ordered (and thus cannot be accessed to add back to the menu themselves). It isn't obvious where they have ended up at this point but they would still be displayed in the All view. Now, with these changes, they will also be hidden.

Probably the best option is to dump all children from a widget if its removed from the menu (best done in both of the identical `reorderTree` functions).

thomasgwatson (approved):
Added some notes; I can make the changes to reorderTree in a different PR

apps/web/src/routes/AllView/AllView.jsx
tibetsprague: no we want those too. good catch

apps/web/src/routes/AllView/AllView.jsx
tibetsprague: another good catch. yeah I think dumping them from the parent makes sense

tibetsprague:
would be great if you can take on the dumping children

tibetsprague:
pushed a fix for the custom views

PR #285, opened 2025-02-19
'my' context and user settings in mobile (LEJ)

lorenjohnson:
@thomasgwatson there is a lot here, but most all of it good. But it may be hard to catch-up to and I'm sorry about that, but hopefully not. Take a look through and ask questions / shoot holes and such. There are definitely some things completed, fixed, and newly broken here (mostly on purpose). I worry it will be hard to see through to the good here as there is still very much some modeling being worked out here, but I do think this moves things a long quite a ways. 

I need to zoom out for an hour to regain some perspective before I can make more concrete notes, but would prefer if we sync at that point and we walk through it together with an eye on next steps.

---

Some remaining work on my mind for check-ing about coming-up towards release, not necessarily strictly related to this PR but to the state of nav/linking and our Mobile App as we approach release:

- widgetUrl generation for /my/stream ... needs attention
- In this branch I've gotten rid of the "Profile / My Face" tab bar item entirely and put it at the top of the groups/context menu like on Web. which I think is the right call. Is there any other shuffling that we need/want to do before release in terms of tabs and top items? Or is this set going to work? 
- I consider this mostly unreleasable without us coming-up with a solution to seeing the names of the groups in the ContextSwitchMenu. Thinking just getting a Long Press event that doesn't propagate to a Press and pops up an ItemSelectorModal with all the groups and a heading that say Select Group (and the standard modal X if they want to cancel). Basically the same box a user sees now when selecting a Group when creating a Post.
- We still need to fill-in some missing routes that I was holding off on (mostly I think EVERYWHERE/post/:postId). I do think on final pass I'm going to consolidate a lot of the linking index table into common scenarios which is why I was holding off. 
- Styling of DrawerMenu (both ContextSwitchMeny and ContextMenu) styling is for me a bit unresolved when it comes to the "safe areas" outside of the content area. Do we want the color of the menu to extend through the safe areas? We can do that. I tried it though and it didn't look quite right either...
- ask-and-offer -> request-and-offers status? 
- There was a comment about Group Explorer going away maybe but I still see it on the latest Web... Tell me more?
- Currently My Profile goes to a Modal with the Member Profile view which includes the feed and such. That's fine. But I think we should add (add back because I think it was there at some point) a Screen called "My Profile" and have it route there within the Home Tab stack so it doesn't feel as much like a "oh by the way" here's your profile in a modal.
- Styling for ContextMenu still needs attention, which I'm sure you are tracking. In particular I'd like to see some of the text get bigger and I'm in question about the indentation for items under headers. It feels like it shouldn't be there, and a bit "expensive" on the limited Mobile screen size. Also the boxy widget items (children?) like for Settings. Those are asking for something.
- Staging performance was pretty laggy and sometimes out-of-service in the last couple weeks. I switched off of it back to my local dev server for my own dev/testing workflow. I am concerned about this for Mobile testing next week. For a complication of reasons I feel like a laggy backend for some reason disproportionately badly impacts Mobile UX... But it's just testing. Will be curious what things will feel like in Production. 
- Curious about the ">" on the Group Header within ContextMenu. I kinda feel like it shouldn't be there and/or the whole header should nav to where it does.
- Are the User and Group Settings areas going to get tightened up? They don't look great in mobile size layout (buttons are weird, etc)
- HyloEditor in the case of @ mentions doesn't auto expand to the height of the selection list. It is workable as is but I think before this was solved, and seems to not be anymore. I don't know if I'll get a chance to look at it before release, probably not, so just mentioning it here.
- We need to decide soon what we want to do, or not, with having a back button where the hamburger menu is. I still suspect someone will veto it not being there, and I'm not sure of my own opinion though... Would like to discuss and weigh options.
- I am a little shakey about our use of openUrl as the standard for ContextMenu in combination with use of navigation.navigate, though after settling some things it seems to be working fine. Let's consider making ContextMenu use all openUrl(path) instead of navigation.navigate.

---
- There is a bug in New Message such that it asks to "confirm leave" when you try and send an initial message, need to create a state flag to say that we're now submitting so leave us alone.


apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -62,23 +56,22 @@ export default function ContextMenu () {
...
   const [{ currentGroup }] = useCurrentGroup()
 
-  const title = widget.title
+  const title = t(widget.title)
   const url = makeWidgetUrl({ widget, rootPath, groupSlug })
-
   const handleWidgetPress = widget => {
-    const context = currentGroup?.isContextGroup ? currentGroup?.slug : 'groups'
-    const linkingPath = makeWidgetUrl({ widget, rootPath, groupSlug: currentGroup?.isContextGroup ? null : currentGroup?.slug, context })
+    const linkingPath = makeWidgetUrl({ widget, rootPath, groupSlug: currentGroup?.slug })
```
thomasgwatson, line 70: Ok, I need to bookmark this

apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -62,23 +56,22 @@ export default function ContextMenu () {
...
+function MenuItem ({ widget, groupSlug, rootPath, group }) {
   const { t } = useTranslation()
   const { listItems, loading } = useContextWidgetChildren({ widget, groupSlug })
   const logout = useLogout()
   const hasResponsibility = useHasResponsibility({ forCurrentGroup: true, forCurrentUser: true })
   const canAdmin = hasResponsibility(RESP_ADMINISTRATION)
   const [{ currentGroup }] = useCurrentGroup()
 
-  const title = widget.title
+  const title = t(widget.title)
```
thomasgwatson, line 67: Need to bookmark this

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -3,23 +3,24 @@ import { Animated, TouchableOpacity, Text, StyleSheet, View, FlatList } from 're
...
-import { openURL } from 'hooks/useOpenURL'
 
 export default function ContextSwitchMenu () {
+  const changeToGroup = useChangeToGroup()
   const [{ currentUser }] = useCurrentUser()
   const [{ currentGroup }] = useCurrentGroup()
-  const changeToGroup = useChangeToGroup()
-  const myGroups = [PUBLIC_GROUP].concat(sortBy('name', map(m => m.group, currentUser.memberships)))
+  const { myContext, publicContext } = useContextGroups()
+  const myGroups = [myContext, publicContext].concat(
```
thomasgwatson, line 18: Wait, are you putting another way to access the my context here?

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -48,7 +50,7 @@ export default function ContextSwitchMenu () {
           </View>
         </TouchableOpacity>
         <TouchableOpacity
-          onPress={() => Intercom.present()} // Needs to open some creation dialog?
```
thomasgwatson, line 51: yes, whoopsie, this was meant to be on the Touchable above

apps/mobile/src/components/DrawerMenu/DrawerMenu.js
```diff
@@ -8,9 +8,10 @@ import ContextSwitchMenu from 'components/ContextSwitchMenu/ContextSwitchMenu'
 
 export default function DrawerMenu (props) {
   const insets = useSafeAreaInsets()
-  const { myHome, groupSlug } = useRouteParams()
+  const { groupSlug } = useRouteParams()
+  const [{ fetching }] = useCurrentGroup({ setToGroupSlug: groupSlug })
 
-  useCurrentGroup({ setToGroupSlug: groupSlug })
+  if (fetching) return null
```
thomasgwatson, line 14: Won't this hide the ContextSwitchMenu while you select a group and it is fetching the full group details?

lorenjohnson: Indeed it will, good catch. Fixed.

apps/mobile/src/util/navigation.js
```diff
@@ -49,12 +45,10 @@ export function baseUrl ({
     return viewUrl(view, { context, customViewId, defaultUrl, groupSlug })
   } else if (groupSlug) {
     return groupUrl(groupSlug)
-  } else if (context === ALL_GROUPS_CONTEXT_SLUG) {
```
thomasgwatson, line 52: Bookmarking this

apps/mobile/src/components/ContextMenu/ContextMenu.js
thomasgwatson: Ok, so we are now bringing useTranslate into the packages. And otherwise we've been resolving the title in the presenters. Why now are we bringing t back here?

apps/mobile/src/util/navigation.js
thomasgwatson: ok I suspect this will get us into trouble.

Some widgets will be passed in with an `all` context.

So their paths won't render correctly and they won't be caught by the linking table.

We can change that in the staticContextGroups, by editing the static widgets...

So then their paths are just substituting `my` in the place of `all` and the path still got us to the Stream screen, then the query will fail, as far as I can tell from this changeset, because there will be nothing indicating to the stream query that it will match with the backend's expectations for the query

If we change this backend line

`onlyMyGroups: context === 'all',` to something like 
`onlyMyGroups: context === 'all' || context === 'my', then we are probably safe.

thomasgwatson:
Ok, I kinda get the feeling that this PR is at least partially about removing `/all`. We have certainly talked about that and Tibet did a trial PR for the web/backend of this but we decided it was too risky at the moment and bailed on it.

So I'm leery of trying to follow through on a similar effort for mobile.

I have to checkout of my airbnb and drive seven hours, so unfortunately won't have much to update you on while you sleep

apps/mobile/src/components/ContextMenu/ContextMenu.js
lorenjohnson: I am glad you asked, this was done because in the current presentations of widget these titles don't get translated and this was a force. I was gonna ask you about this, but for the moment this renders the expect UI.

apps/mobile/src/util/navigation.js
lorenjohnson: I agree with your suspicions, but not yet with the solution. Will
loop back more.

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
lorenjohnson: yes see GroupPresenter

lorenjohnson:
This doesn't do away with the "/all" routing, it simply reserves the word "context" as something about UI state on one end, and a special stream query variable on the other end, decoupling the two.

tibetsprague:
to answer some questions that may not have been answered. ask-and-offer has officially transitioned to requests-and-offers

tibetsprague:
Group Explorer is not going away. The Group's "Explore" page is :)

tibetsprague:
On the web for mobile devices there is no hamburger menu anymore, just a back button that takes you back to the Nav + context menu. we should match the app to the web as much as possible. but maybe im missing the need for something different?

tibetsprague:
there is a lot of UI work still to do on web, so yes buttons will look weird right now.

tibetsprague:
Agree that there should be a My Profile page that display the profile in its own route (as there is on web)

thomasgwatson:
> In this branch I've gotten rid of the "Profile / My Face" tab bar item entirely and put it at the top of the groups/context menu like on Web. which I think is the right call. Is there any other shuffling that we need/want to do before release in terms of tabs and top items? Or is this set going to work?

It is where it is meant to be based on the last product/design discussions.

This is the closest we get from the figma. It represents a design that is strictly speaking, not possible, since the bottom tabs and the drawer never show up like this (the drawer always goes over the main screens/bottom tabs). Also, the existing 'home' tab isn't accounted for in the designs;

Screenshot 2025-02-21 at 7 53 42 AM

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
thomasgwatson: no, i mainly meant as a piece of the UI, which you have clarified elsewhere.

thomasgwatson:
- Safe Area; complicated by the presence of Group Banners as well, since they'll be a range of colors. Not sure what the best course of action is there.
- The route for a member's profile is just dropping through to the modal version. Is that screen still relevant? It might just need to be pointed at the right screen, and also then, whats the purpose the of screen its currently pointing at?
- Yes, there is more work to do with styling the context menu, special behavior for the members widget, etc etc...
- ContextMenu have five uses of navigation.navigate; four of them could be replaced with an existing/meaningful route, one can't really be replaced (Group Welcome).

apps/mobile/src/components/ModerationList/ModerationList.js
```diff
@@ -96,6 +96,8 @@ export default function ModerationList ({ forGroup, header, scrollRef, streamTyp
     }
   }, [hasMore, pending])
 
+  const handleDecisionOptionsOnChange = () => navigate('Stream', { streamType: 'proposal' })
```
thomasgwatson, line 99: This is going to be redone anyway once this branch gets Tibet's context widget additions

apps/mobile/i18n.js
```diff
@@ -1,4 +1,4 @@
-import i18next from 'i18next'
+import i18n from '@hylo/shared/i18n'
```
lorenjohnson, line 1: It's not expected that this i18n as shared module stuff is coming in here, it is already merged and once it went to dev I merged it up to your branch with this is based upon... Anyway, you never got to review this bit so no harm but this PR was a bit shorter without this extra.

lorenjohnson:
> @brodeur definitely has opinions (and designs) on which buttons should be in the bottom bar. i cant track it down right now. but i think search wants to be down there too?

Yes, search is still down there. I misspoke in an earlier remark as if it was gone. We can review all of this next week. The moves now are the quick and easy (free ones) to bring what is there into coherency.

lorenjohnson:
> > In this branch I've gotten rid of the "Profile / My Face" tab bar item entirely and put it at the top of the groups/context menu like on Web. which I think is the right call. Is there any other shuffling that we need/want to do before release in terms of tabs and top items? Or is this set going to work?
> 
> It is where it is meant to be based on the last product/design discussions.
> 
> This is the closest we get from the figma. It represents a design that is strictly speaking, not possible, since the bottom tabs and the drawer never show up like this (the drawer always goes over the main screens/bottom tabs). Also, the existing 'home' tab isn't accounted for in the designs;
> 
> Screenshot 2025-02-21 at 7 53 42 AM

Yes the unresolved issues I see in this design, regardless of how we'd implement which shouldn't too bad... is that I don't see a clear way to navigate back / between screens in a stack. The browser has back buttons and a familiar feeling history stack. Mobile is a bit different and we manage it more. I suspect trying to implement this verbatim would reveal a few strange unanswered questions of that order, but maybe it would be fine? Other than that, yes, having the sidebar there all the time doesn't feel right for Mobile. I think on all but larger devices it would be too confining for the screen real estate. 

Otherwise, in taking this lead I'd be happy to put the notifications bell at the bottom right, though I am not yet sure in the current design it would be better there. It might be though. Things at the bottom are a good thing due to thumb proximity. I also like the idea of free-ing up a bit of top bar space to make titles of things (like the group name) more visible.

lorenjohnson:
Some quick and dirty illustrations of current state of things for those of you who don't get to booting the app :)

Simulator Screenshot - iPhone 15 Pro Max - 2025-02-20 at 23 29 32
Simulator Screenshot - iPhone 15 Pro Max - 2025-02-20 at 23 29 56
Simulator Screenshot - iPhone 15 Pro Max - 2025-02-20 at 23 31 09

packages/hooks/useStreamQueryVariables.js
```diff
@@ -30,7 +29,7 @@ export default function useStreamQueryVariables ({
     childPostInclusion: currentUser?.settings?.streamChildPosts || 'yes',
     collectionToFilterOut: null,
     context,
-    createdBy: context === MY_CONTEXT_SLUG || myHome === 'posts'
+    createdBy: myHome === 'posts'
```
lorenjohnson, line 32: Good catch. I tend to forget that boolean values pass through

packages/hooks/useStreamQueryVariables.js
```diff
@@ -48,7 +47,12 @@ export default function useStreamQueryVariables ({
       ? (timeframe === 'future' ? 'asc' : 'desc')
       : null,
     search: null,
-    slug: !isContextGroupSlug(forGroup?.slug) ? forGroup?.slug : null,
+    // We just can't get away from some sort of jank until /all is removed from the whole repo
```
lorenjohnson, line 50: While I agree, and believe that jank is in fact entirely contained to the getStateFromPath changes that redirect things (and the paired change to create this temporary "/my/stream" route, which is a good call, this change here is not correct or necessary. The backend knows nothing of an 'all', 'public', or 'my' slug and the previous logic was correct. This might be different due to some additional indirection on "slug" within the query maker on Web, but not here. What am I missing, and why did you make this change here?

apps/mobile/src/navigation/linking/index.js
```diff
@@ -76,6 +76,7 @@ export const routingConfig = {
   '/:context(my)/:myHome(interactions)':                                  `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
   '/:context(my)/:myHome(mentions)':                                      `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
   '/:context(my)/:myHome(posts)':                                         `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
+  '/:context(my)/:myHome(stream)':                                        `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
```
lorenjohnson, line 79: Good call! As this is a temporary and integral route, and not something we have on our source-routing on Web, please put a comment above it noting that and why it is here. Also was not expecting that you'd put it under the "myHome" named param. Is there a reason we need to do that here? myHome is a bit of a special case with logic attached downstream so I was thinking we wouldn't name it or give it a unique janky temporary name so we don't end-up with myHome logic embedded somewhere that breaks later? Not too worried, just sayin.

apps/mobile/ios/Podfile.lock
```diff
@@ -2484,7 +2484,7 @@ SPEC CHECKSUMS:
   Mixpanel-swift: 2192b9a24cf41b870749e4e0e10fbc9822bb2b4a
   MixpanelReactNative: 76314cf184a17d493e2dd8a3e367c81d91361eea
   OneSignalXCFramework: f5b2a3c4f130e4c910ead7bb25bed7455e976fbf
-  RCT-Folly: bf5c0376ffe4dd2cf438dcf86db385df9fdce648
+  RCT-Folly: 84578c8756030547307e4572ab1947de1685c599
```
lorenjohnson, line 2487: I think if one or both of us cleared our pod cache we might be able to stop the ping pong here, though it doesn't hurt me really and needs no attention. Just silly.

apps/mobile/src/navigation/linking/index.js
thomasgwatson: If anything, I think this is "skating to where the puck is going"; where web will (could) end up anyway. Already reverted the :myHome-ing of this before I saw this comment

lorenjohnson: Well for web though we came to the conclusion, at least for now that this route will be `my/groups/*` which I think is good. So the comment really will help. There are no other routes in the table that are internal or not mapped to Web, and we'd like to keep it that way.

lorenjohnson: Oh sorry, I took your remark to be saying that your work was done here:) I see now you were speaking of the query variables slug thing. Yay. Ok, sail on

apps/mobile/src/navigation/linking/index.js
```diff
@@ -94,6 +93,8 @@ export const routingConfig = {
   '/:context(my)/:streamType(requests-and-offers)':                       `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
   '/:context(my)/:streamType(resources)':                                 `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
   '/:context(my)/map':                                                    `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Map`,
+  '/:context(my)/stream':                                        `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
```
lorenjohnson, line 96: I am picky about the column spacing

lorenjohnson:
> On the web for mobile devices there is no hamburger menu anymore, just a back button that takes you back to the Nav + context menu. we should match the app to the web as much as possible. but maybe im missing the need for something different?

We can make our hamburger a back button icon instead, but yes, Native is different than Web in the ways I described, which are worth considering. In the Web implementation in most all cases there is one page, then a modal that pops-up for detail. We can treat all our detail like things (e.g. PostDetail, and Member) in the same way and given them and X instead of a back button and make the hamburger a back button. Previously the hamburger would turn into a back button when you were more than one screen into the stack, we can also bring that back. 

For the most part nothing much has changed at this level in the UI though, nor been given attention, because the re-write that has happened in these last months is all well below this level. I do think once we get to it that using full screen Modals for like PostDetails and the like (so the screens with X buttons) and retiring the hamburger menu (I guess that is so 2012 now), should work fine. 

We just need to keep in mind that this means that things like going to a Member while on a Post Detail page means that you will have another modal on top of that, which is fine, and in effect not different than what happens now. But stack-based native navigation gives you animations like you're going back in history vs opening and closing things... Hope that is making sense. The difference is a little more than subtle, and not purely aesthetic, but also not a big deal. I just want, and have wanted for a bit, to feel like these mobile/native specific things are given some consideration because all the implementation in Native will make you aware of them if you want to be or not, and it is easy to start forcing designs which work against the grain of what otherwise "just works" and gives a solid and expected mobile UX.

Hope that is coming across right. This topic of the Mobile design has always to me felt a little like something was supposed to be happening and somehow obvious that didn't feel to me was otherwise being carefully looked at in its implications. I don't feel that way much at the moment, but a bit on guard, and the amount of things we're doing at once right now has once again--I believe--left the more nuanced considerations of what makes a good Mobile UX perhaps a bit still under considered. And again, that is fine, but I am hesitant to push hard and fast on any Mobile redesign if we really don't have to, and can this round give some of what I'm saying proper consideration as we go. And maybe that is exactly the plan. I like to think so. 

_Ps. Some other common mobile app interactions which are by default implemented in a React Navigation stack and are not so common on web: also press on the tab icon to return to the top of the stack at anytime, or swipe to go back.

lorenjohnson:
All I want for Christmas is to be able to see the names of my Group on Mobile. As long as we come up with a tidy way for that to be possible by release, I will call this a success :)

lorenjohnson:
> Agree that there should be a My Profile page that display the profile in its own route (as there is on web)

Yep--by default opens it was opening in a modal, and now it opens within the tab, that was the only difference here, all the routing was already in place. It is in fact just a link to the member page by id in both cases, vs having an explicit page or screen for "My Profile", and that is good by me.

apps/mobile/src/navigation/linking/index.js
```diff
@@ -99,7 +99,7 @@ export const routingConfig = {
   // /all  routes -- NOTE: the "all" context is re-mapped to "my" by addParamsToScreenPath in getStateFromPath)
   '/:context(all)/map':                                                   `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Map`,
   // TODO routing: commenting this out and letting the modal catch-all handle it. Thoughts?
```
lorenjohnson, line 101: This comment can now be removed.

tibetsprague:
oh yes, I do think I agree that not seeing the names is going to be not ideal. but keep forgetting to bring that to @brodeur in a product call. I think that having them appear when you scroll through global nav could work well

tibetsprague:
noting that one failing of our process and our dev work in general is not ramping Aaron up to understanding the nuances and differences of mobile app UX. he really wants to get in there and understand and work on it, and we just have not dont that yet. I feel you Loren, and that's on us.

thomasgwatson:
> oh yes, I do think I agree that not seeing the names is going to be not ideal. but keep forgetting to bring that to @brodeur in a product call. I think that having them appear when you scroll through global nav could work well

We did discuss this briefly in a product call and the general idea of the long-press/scroll-press leading to the names showing has made the full circuit of product and dev. We just need to implement it

lorenjohnson:
> noting that one failing of our process and our dev work in general is not ramping @brodeur up to understanding the nuances and differences of our mobile app UX and how it is different from there web. he really wants to get in there and understand it and work on it, and we just have not dont that yet. I feel you Loren, and that's on us.

Sounds good. To be clear I am not necessarily meaning to hold up our particular Mobile UX as-is or ever, but general Mobile UX in difference to Web. Less anything I think mysterious, just something that I think warrants specific attention and a level of consideration that I simply think we have not had time to give it usually, and maybe still somewhat. The sorta points I bring up are only because these differences become more clear when "down in it" in the implementation where there are sometimes surprising / unexpected things to consider when coming from a browser/Web perspective.

PR #286, opened 2025-02-19
Hide empty container widget titles

tibetsprague:
So we dont have floating titles of widgets with no content, which is confusing. 

not 100% sure im catching all the cases here


thomasgwatson:
This is being addressed elsewhere, I'd just bail on this PR

PR #288, opened 2025-02-20
List day calendar view

KevinTriplett:
Replaces PR #267 

This PR beautifully displays multi-day events, removes the week view, and lists posts when in day mode.

I'd merge this, but discussion / approval is needed [for the query revisions here](https://github.com/Hylozoic/hylo/pull/288/commits/e20b4623faf4159a484a1c9c3437efeeb1c84ea6#diff-eba6e817ffce5299794eb3cda326ac0a23b4cb8913f7dfe1488246d0d88d4a3f).


KevinTriplett:
Here are two view, month and day.

Screenshot 2025-02-19 at 3 02 59 PM

Screenshot 2025-02-19 at 7 53 23 PM

apps/backend/api/services/Search/util.js
```diff
@@ -75,14 +75,17 @@ export const filterAndSortPosts = curry((opts, q) => {
     })
   }
 
-  if (afterTime) {
+  if (afterTime && beforeTime) {
+    q.where(q2 =>
+      q2.where('posts.start_time', '<', beforeTime)
+      .andWhere('posts.end_time', '>', afterTime)
```
tibetsprague, line 81: I think it should be >= afterTime since thats what we do below

apps/web/src/components/Calendar/body/calendar-body.tsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react'
 import { useCalendarContext } from '../calendar-context'
 import CalendarBodyDay from './day/calendar-body-day'
-import CalendarBodyWeek from './week/calendar-body-week'
```
tibetsprague, line 4: agree we dont need this now, but maybe create a ticket to bring it back eventually?

apps/web/src/components/Calendar/body/month/calendar-body-month.tsx
```diff
@@ -32,6 +32,8 @@ export default function CalendarBodyMonth () {
     (event) =>
       interval.contains(DateTime.fromJSDate(event.start)) ||
       interval.contains(DateTime.fromJSDate(event.end))
+  ).sort(
+    (a, b) => a.multiday ? -1 : a.start.getTime() - b.start.getTime()
```
tibetsprague, line 36: within multi-day events dont we still need to sort them amongst each other?

apps/web/src/routes/Stream/Stream.js
```diff
@@ -149,16 +149,14 @@ export default function Stream (props) {
       const luxonDate = DateTime.fromJSDate(calendarDate)
       switch (calendarMode) {
         case 'month':
+        case 'day':
           params.afterTime = luxonDate.startOf('month').startOf('week', { useLocaleWeeks: true }).startOf('day').toISO()
           params.beforeTime = luxonDate.endOf('month').endOf('week', { useLocaleWeeks: true }).endOf('day').toISO()
           break
         case 'week':
```
tibetsprague, line 156: didnt we get rid of week? or leaving this in for the future?

tibetsprague (changes requested):
good stuff! a couple little things

KevinTriplett:
Added week mode:

Screenshot 2025-02-20 at 3 44 34 PM

apps/backend/api/services/Search/util.js
```diff
@@ -75,14 +75,17 @@ export const filterAndSortPosts = curry((opts, q) => {
     })
   }
 
-  if (afterTime) {
+  if (afterTime && beforeTime) {
```
tibetsprague, line 78: actually i just realized, might it not be good enough to just have the afterTime and beforeTime conditionals separate from each other? wouldnt that cover the case that both are set? also im trying to understand why the conditions in afterTime & beforeTime are different from the cases where only one is set. Seems like they should be the same, and then we only need to the two separate conditions

apps/backend/api/services/Search/util.js
```diff
@@ -75,14 +75,17 @@ export const filterAndSortPosts = curry((opts, q) => {
...
+    q.where(q2 =>
+      q2.where('posts.start_time', '<', beforeTime)
+      .andWhere('posts.end_time', '>=', afterTime)
+    )
+  } else if (afterTime) {
     q.where(q2 =>
       q2.where('posts.start_time', '>=', afterTime)
```
tibetsprague, line 85: you're right, i think only needs to be end_time >= afterTime, dont need to worry about the start tike condition. unless... what if we have a post with a start time but no end time? i dont think we allow that right now though

apps/backend/api/services/Search/util.js
```diff
@@ -75,14 +75,17 @@ export const filterAndSortPosts = curry((opts, q) => {
...
       .orWhere('posts.end_time', '>=', afterTime)
     )
-  }
-
-  if (beforeTime) {
+  } else if (beforeTime) {
     q.where(q2 =>
       q2.where('posts.start_time', '<', beforeTime)
       .andWhere('posts.end_time', '<', beforeTime)
```
tibetsprague, line 91: this one is a judgement call, and you made a different judgement above when both are set. does beforeTime mean the whole event needs to be before that time? I said yes, but you said no. I actually think you are right again. so we can definitely simplify this whole thing into two separate condtions using the ones you used.

tibetsprague (changes requested):

apps/backend/api/services/Search/util.js
```diff
@@ -75,21 +76,28 @@ export const filterAndSortPosts = curry((opts, q) => {
     })
   }
 
-  if (afterTime && beforeTime) {
+  if (calendarViewMode) {
```
tibetsprague, line 79: hmmm this isn't what I was suggesting. I was suggesting we dont need a special condition for calendars, you can just have the if (afterTime) become a separate condition from if (beforeTime), then we should be able to handle the calendar queries without any special conditions. no?

tibetsprague: also, does this work? if you add a new query param on the backend then it would need to be added to the schema.graphql and probably in several other places too. I think we should be fine with just beforeTime and afterTime

apps/backend/api/services/Search/util.js
tibetsprague: i guess im not communicating this clearly enough. i still don't see why we need three conditions here: one for afterTime & beforeTime, one for just afterTime and one for just beforeTime. Can't we just have two separate if statements for afterTime and beforeTime, and that would handle it? like this:

if (afterTime) {
      q.where(q2 =>
        q2.where('posts.endTime', '>=', afterTime)
      )
  }

if (beforeTime) {
  q.where(q2 =>
      q2.where('posts.end_time', '<', beforeTime)
    )
}

tibetsprague (changes requested):

KevinTriplett:
Thank you for persevering, I looked at our two queries closely and found a way to collapse the if statements. But I have to respectfully disagree with your proposal, it's _almost_ correct.

This is the query calendar needs (I know* this works, I've done it on two previous scheduling apps that had calendar displays):

> `post.start_time < beforeTime AND post.end_time >= afterTime`

This is the query you propose:

> `posts.end_time < beforeTime AND posts.endTime >= afterTime`

They are very similar with one mucho importante differencia**. But they did show that I could roll the three tests into two tests, as you requested ( in [this commit](https://github.com/Hylozoic/hylo/pull/288/commits/2b17521334bda0a3ebc0507b89027452354110f7)):

````
  if (afterTime) {
    q.where(q2 =>
      q2.where('posts.end_time', '>=', afterTime)
    )
  }
  
  if (beforeTime) {
      // if afterTime also supplied, this is a calendar view
      // so get posts that intersect the date range
    q.where(q2 =>
      q2.where(afterTime ? 'posts.start_time' : 'posts.end_time', '<', beforeTime)
    )
  }
````

Sorry, I believe this is the best we can do.

*Unless I'm being totally dense, in which case I'll resign as a volunteer dev, you don't need imbeciles on the team ;-)

**Note that `posts.end_time < beforeTime AND posts.endTime >= afterTime` would only fetch posts that end during the calendar period being displayed.

**Comment by tibetsprague** [comment]
ok nice, i think i'm good with this version. However, I want to think through it one last time. The only time we use beforeTime is in card/list view, when showing Past events, so if we go with posts.end_time', '<', beforeTime then that means we only show events that have fully ended, not ones that are currently happening. Do we want that? 🤔 i actually like the idea of showing currently happening events both in the past events view, and in the upcoming events view, since it is true that the event is both past (already started), and upcoming (still going). This means we could remove the afterTime condition inside the beforeTime condition and just always use 

```
if (beforeTime) {
      // so get posts that intersect the date range
    q.where(q2 =>
      q2.where('posts.start_time', '<', beforeTime)
    )
  }
```

Does that make sense to you?

KevinTriplett:
> Does that make sense to you?

I like where you're going. Let's look at both queries:

`posts.endTime >= afterTime` <== retrieves ongoing and future posts

`posts.start_time < beforeTime` <== retrieves past and ongoing posts

(In both cases, `ORDER BY posts.start_time ASC`, per Stream.js)

This works for calendar view nicely. In that case:

````
  // NB: ongoing posts are returned in both cases
  // NB: calendarView needs WHERE post.start_time < beforeTime AND post.end_time >= afterTime
  if (afterTime) {
    q.where(q2 =>
      q2.where('posts.end_time', '>=', afterTime)
    )
  }
  
  if (beforeTime) {
    q.where(q2 =>
      q2.where('posts.start_time', '<', beforeTime)
    )
  }
````

I'm okay with showing ongoing posts in both views. Personally, I expect past posts to not include ongoing posts but it would be worth the experiment. I commit this change and we can easily revert back if we get complaints. :-)

tibetsprague (approved):

tibetsprague:
whoo hoo!

PR #292, opened 2025-02-22
Next-up: Fixes, Exploring/addressing key UX issues, and code-level tidying-up

lorenjohnson:
Some things I'd like to get done in this branch:

- [x] Make ContextSwitchMenu switch to a new context or group home, not just to the current selected view. I think somewhere in Widget-land there is / should be a way for me to simple "getHome" for a particular group/context so I know the right default place to direct to?
- [x] Post big-deltas code tidy-up alignment (like naming of graphql things, etc)
- [ ] Test/Fix SignUp
- [ ] Test/Fix Create Group
- [ ] Test/Fix JoinGroup (from auth and non-auth)
- [ ] Test/Fix returnToOnAuth
- [ ] Urql: Fix MesageThread subscription flashing/resetting on new message arriving from updates subscription
- [ ] Urql: Fix Reaction Deletion for Comment in PostDetails
- [ ] Improve keyboard handling on the CommentEditor, after removal of KeyboardAccessoryView we probably want adopt another ScrollView-based keyboard avoiding library to make the keyboard behaviour less clunky. At least turn-off iOS "interaction" mode keyboard slide as the TextInput doesn't keep up with that animation without said special library/handling. 
- [ ] Add CurrentUserPresenter, maybe. It would have some additional methods on it like `getAllContexts()` which returns all groups as well as the two context groups. Will also look into the binding of currentUser in any GroupPresenter or ContextWidgetPresenter logic to see if there is anythning to migrate here. currentUser is a good candidate for a presenter as we already always and only access currentUser through the useCurrentUser hook, but there actually isn't hardly any transformations we do on currentUser data currently, which is a good thing. However, the places we bind it may tell another story that can be clarified by reversing the dependency... It's a brief exploration.
---

Other areas I will be looking into shortly, but probably in a separate PR/s so they can be evaluated discussed independent of the more objective tasks above:

- [ ] ContextMenu styling: I already in an experimental branch updated the styling to align it more with Web, using some regular React Native Stylesheet things, in combination with NativeWind theming for colors. That worked well and was quite quick to whip things into shape. The resulting styling configuration between the style and classNames props was to my eyes quite clear and easy to grok. I would like to do this again here (or likely in a branch off of this for easy evaluation), to both get the menu a bit more usable (currently things are somewhat critically too small IMHO) and explore/introduce the best pattern for styling now that we have NativeWind and as guidance before @brodeur digs in here. 
- [ ] The Group Settings screen is currently accessed either through the gear icon in the group header or through a series of targeted settings areas links coming in through the Widgets. The links to the areas from the Widgets all go to the Group Settings screen but not to the particular settings area described in the link. Fix this and evaluate the gear+settings area+settings menu pattern to see what might be more clear UX-wise. 
- [ ] Try switching PostDetails and Member profile Screens to be Full Screen Modals instead of Stack Screens. This is coming from a discussion with @tibetsprague on the previous PR about aligning Mobile UI patterns with those of Web. This will be more a functional change than styling, and needs to be tried to feel how this is as a Mobile pattern. My suspicion is that it will work just fine, and answer some implicit "Back Button" questions we're soon going to otherwise have.
- [ ] Regardless of conclusions of the "Full Screen Modal for Detail screens" exploration in the item above this, sort-out Hamburger / Back Button UX

---

@thomasgwatson tagging you here as an FYI of some things I'm hoping to address shortly. The above 2 lists will probably get longer, and some items may get dropped off or clarified. If there are any of these items you anticipate getting to yourself, please indicate in some way (maybe just @ mention yourself next to any item), and then I will leave it be for you.


thomasgwatson:
Would be great to see the styling syntax you are settling on.

thomasgwatson:
> The Group Settings screen is currently accessed either through the gear icon in the group header or through a series of targeted settings areas links coming in through the Widgets. The links to the areas from the Widgets all go to the Group Settings screen but not to the particular settings area described in the link. Fix this and evaluate the gear+settings area+settings menu pattern to see what might be more clear UX-wise.

The setup widget is optional, and can be removed. So the gear will need to remain one way or another.

And while we are here: All widget and menu editing is only available on web. At some point we might offer a limited way to tweak the menu but not anytime soon

PR #294, opened 2025-02-23
Revert shared i18n, make translation a component-level responisiblity

lorenjohnson:
- Removes shared i18n instance due to malfunction
- Makes translation a component-level responsibility
- Adds translateTitle to ContextWidgetPresenter and applies it
- Removes `models/EventInvitation` `RESPONSES` and `humanResponse` on Web and applies `EventInvitationPresenter` for those as was already done on Mobile. _It is a gradual transition towards URQL step to move from storing utility methods like this on the Model instead of the presenter, and regardless we should keep the logic in one place._


packages/presenters/ContextWidgetPresenter.js
```diff
@@ -149,11 +144,9 @@ export function findHomeWidget (group) {
   return group.contextWidgets.items.find(w => w.parentId === homeWidget.id)
 }
```
lorenjohnson, line 146: Excuse the noise below, just alphabetizing exported methods.

tibetsprague (approved):

PR #308, opened 2025-02-26
Mobile Signup, clean-ups

lorenjohnson:
chore(mobile): Signup fixes/refactors (WIP). LocationSelector non-modal option for Signup Location. Fixes upload images. Fixes ModalHeader close icon default color. Creation modal and PostEditor clean-ups/linting.


thomasgwatson:
So Create Group works as is, aside from the very last step: navigating to the newly created group.

So this touches on another issue; openURL doesn't seem to impact currentGroup. Or maybe more broadly the linking table doesn't?

The app registered the link but didn't change the currentGroup in response to it. Separately, I used yarn open-link /groups/:groupSlug and that also didn't impact the currentGroup. I checked `getStateFromPath` and it had the right path.

Essentially, I'm wondering if we should include changeToGroup inside of openURL... But then, if the app receives a link externally, it won't care about openURL but instead just respond to where-ever the linking table drops us. 

So I'm trying to figure out at what point we intervene to ensure the group actually gets updated.

I thought we had moved a setToGroupSet check from somewhere lower in the component hierarchy, up to a more central position but now I can't find that. And I don't think just adding an exception to openURL will work for all use-cases. So I'm assuming we need to cross-check routeParams against the currentGroup and sync them up.

I'll try a version of that focusing on AuthNavigator and then can see what you think

apps/mobile/src/screens/PostEditor/PostEditor.js
```diff
@@ -266,7 +266,7 @@ export default function PostEditor (props) {
         : t('Post')
 
     return () => (
-      <View className="border-border bg-background" style={styles.headerContainer}>
```
lorenjohnson, line 269: @thomasgwatson as per these double to single quote conversions: get yer linter going please, and hit me up if you need any tips on getting it working.

apps/mobile/src/screens/CreationOptions/CreationOptions.js
```diff
@@ -6,52 +6,48 @@ import { openURL } from 'hooks/useOpenURL'
...
   const handleCreatePost = () => {
     openURL('/create/post')
   }
-
   const handleCreateGroup = () => {
     openURL('/create/group')
   }
 
   return (
-    <View className='flex-1 bg-background p-4'>
```
lorenjohnson, line 19: It is easy in React Native, not unlike web, to end-up with an extra View wrapper where it is unnecessary. It has little impact, but in my experience it can tend to become a source of unnecessary complexity over time with styling and such. So always confirm you need the extra container, most often "container" containers are unnecessary.

apps/backend/api/graphql/mutations/user.js
```diff
@@ -13,13 +13,18 @@ export const sendEmailVerification = async (_, { email }) => {
     }
 
     const { code, token } = await UserVerificationCode.create(email)
+    const verifyUrl = Frontend.Route.verifyEmail(email, token)
+
+    if (process.env.NODE_ENV === 'development') {
```
lorenjohnson, line 18: Help for SignUp UI debugging

apps/mobile/src/screens/PostEditor/PostEditor.js
thomasgwatson: I feel like I've reset it two separate times now... something is not sticking or persisting on restarts? Getting ridic

lorenjohnson: Hmm... I think you'll find if you open from the root of the monorepo it should "just work" as long as you have the StandardJS extension installed. It also usually works for me when I am opening within an apps directory, but I may have a custom setting in my user workspace to have it see standard in the monorepo deps (not sure, have to check). Let me know what you find.

lorenjohnson: This is the extension I mean: https://marketplace.visualstudio.com/items?itemName=standard.vscode-standard

tibetsprague (approved):
lots here, and i didnt look tooooo closely, but looks good!

apps/mobile/src/hooks/useOpenURL.js
```diff
@@ -25,18 +26,6 @@ export async function openURL (providedPathOrURL, reset, navigation = navigation
   ) {
     const linkingPath = linkingURL.pathname + linkingURL.search
 
-    if (DEBUG) {
```
thomasgwatson, line 28: Can we just comment this out, for future debugging aid?

apps/mobile/src/navigation/linking/index.js
```diff
@@ -37,7 +37,7 @@ export const routingConfig = {
   // Auth & Signup Routes
   '/login':                                                               `${NON_AUTH_ROOT_SCREEN_NAME}/Login`,
   '/reset-password':                                                      `${NON_AUTH_ROOT_SCREEN_NAME}/ForgotPassword`,
-  '/signup/:step(verify-email)':                                          `${NON_AUTH_ROOT_SCREEN_NAME}/Signup/SignupEmailValidation`,
+  // '/signup/:step(verify-email)':                                          `${NON_AUTH_ROOT_SCREEN_NAME}/Signup/SignupEmailValidation`,
```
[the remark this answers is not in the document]
thomasgwatson: Yeah, do we want to keep this or not?

apps/mobile/src/screens/PostEditor/PostEditor.js
thomasgwatson: Ok, I'm seeing that my linting server keeps crashing... I'm cutting away a bunch of extensions to see if that helps but there is some suggestion that cursor might be chugging on large code bases (which I guess we get a lot larger now in monorepo land)

Possibly related cursor issue:

https://github.com/getcursor/cursor/issues/1888

apps/mobile/src/screens/PostEditor/PostEditor.js
```diff
@@ -184,9 +185,8 @@ export default function PostEditor (props) {
       e.preventDefault()
       confirmDiscardChanges({
         onDiscard: () => navigation.dispatch(e.data.action),
-        title: t('Are you sure?'),
-        confirmationMessage: t('If you made changes they will be lost'),
-        t
+        title: 'Are you sure?',
```
thomasgwatson, line 188: Will these be translated elsewhere?

thomasgwatson: Ok, nm, I see it

apps/mobile/src/screens/Stream/Stream.js
```diff
@@ -194,6 +192,7 @@ export default function Stream () {
     updateUserSettings({ changes: { settings: { streamChildPosts: childPostInclusion } } })
   }
 
+  // TODO: Unused. Remove or explain in comment
```
thomasgwatson, line 195: Ahh interesting. This was for a stream control toggle. Looks like its being handled elsewhere now, since it works as expected on dev. So we can nuke this.

packages/presenters/GroupPresenter.js
```diff
@@ -51,13 +51,15 @@ function shouldWelcomeResolver (group, currentUser) {
 
 function contextWidgetsResolver (group, currentUser) {
   if (isContextGroupSlug(group.slug)) {
-    return getStaticMenuWidgets({
-      isPublicContext: group.slug === PUBLIC_CONTEXT_SLUG,
-      isMyContext: group.slug === MY_CONTEXT_SLUG,
-      profileUrl: profileUrlResolver(currentUser)
-    })
+    return [
```
thomasgwatson, line 54: I don't get what is going on here?

Is this just something to handle a contextGroup having its contextWidgets resolved several times?

thomasgwatson: Oh, this is to have a home widget?

thomasgwatson (approved):
lgtm

apps/mobile/src/screens/PostEditor/PostEditor.js
lorenjohnson: I'm not having trouble in any of the apps linting at the moment. The only setting I have in my user workspace that could be related:

```
    "standard": {
        "enableGlobally": true
    },
```
🤷

apps/mobile/src/components/ImagePicker/ImagePicker.js
```diff
@@ -117,7 +121,9 @@ export async function showImagePicker ({
 
             onChoice && onChoice({ local: asset.uri, remote: null })
 
-            const { payload, error } = await upload(type, id, file)
+            const response = await upload(type, id, file)
+            console.log('!!!!', response)
```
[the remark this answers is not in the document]
lorenjohnson: remove in another pr :)

apps/mobile/src/hooks/useOpenURL.js
lorenjohnson: The pathMatcher is now part of the routeParams, so is already logged in the matched state object which was the primary thing achieved by this extra debugging.

apps/mobile/src/screens/Stream/Stream.js
lorenjohnson: Ok, next time you see it do that, otherwise I will try and remember when I see it next :)

packages/presenters/GroupPresenter.js
lorenjohnson: Yes, I think there are comments about this in the code. For mobile there is an added static homeWidget for the static my and public groups.

PR #309, opened 2025-02-26
Fix datetime translations

KevinTriplett:
This fixes some of the translation problems with posts.

Tight review of i18n and TextHelpers, to make sure notifications are okay with the changes.

* for i18n, we need to load the namespaces in order to use the translation outside of react components
* for TextHelpers, it was a bit of a mess when luxon could solve the problem AND provide translation win!
* BUT -- there's not an easy way to remove the "ago" from all the translations. And "short" may not be good enough for English everywhere.


KevinTriplett:
This commit "add setLocale as needed" has touched a number of files. I made an executive decision to change the format for locales that don't use English ordering of date (month-day-year) -- so check especially Invite and Join request formatted dates, to make sure it doesn't mess up any automation.

KevinTriplett:
I removed reviews, since translation is probably not a priority.

packages/shared/src/TextHelpers.js
```diff
@@ -1,3 +1,5 @@
+import i18n from '../../../apps/web/src/i18n.mjs'
```
tibetsprague, line 1: this cant be right. we probably should just move this file to shared? hmm...

packages/shared/src/TextHelpers.js
```diff
@@ -1,3 +1,5 @@
...
+import { getLocaleAsString } from '../../../apps/web/src/components/Calendar/calendar-util'
```
tibetsprague, line 2: we shouldnt be loading stuff from web in shared, so move this here?

tibetsprague (changes requested):
great! generally looks good. will require lots of testing across front and backend since TextHelper is used in both places

packages/shared/src/TextHelpers.js
KevinTriplett: After the launch, let's re-visit this potential need for re-organization.

I want to create DateTimeHelper functions, there are many repeating patterns and we can change the library more easily if they're centralized.

KevinTriplett:
closing this one as it's too old to merge into dev

PR #312, opened 2025-02-27
New Group Welcome Page

tibetsprague:
Added in the settings as a rich text field.
Enabled inline images in HyloEditor! Only displayed/used for welcome page right now

Also add website URL to group settings.


lorenjohnson (approved):
LGTM 👍🏼 Made notes about this on the related before release todo item at at the bottom of this list:  https://github.com/Hylozoic/hylo/discussions/323

apps/web/src/routes/GroupWelcomePage/GroupWelcomePage.jsx
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect } from 'react'
+import { useSelector } from 'react-redux'
+import HyloHTML from 'components/HyloHTML'
+import { useViewHeader } from 'contexts/ViewHeaderContext'
+import useRouteParams from 'hooks/useRouteParams'
+import getGroupForSlug from 'store/selectors/getGroupForSlug'
+
+function GroupWelcomePage () {
```
thomasgwatson, line 8: Checked this path with a non-member of a group and it throws error 'group is null'. Forward to the public facing about page or not found?

PR #320, opened 2025-02-28
initialURL and returnToOnAuth link handling

lorenjohnson:
- [x] Test/Fix returnToOnAuth (forwarding to an auth route after login when arriving in app from a link but not auth'd)
- [x] Test/Fix JoinGroup (from auth and non-auth)
- [x] Fixes Keyboard handling in GroupWelcome questions (adds react-native-keyboard-manager)
- [x] Refactors GroupWelcomeFlow to use zustand
- [x] Some sign-up tweaks, new user arrival fixes
- [x] Fixes issue with useCurrentGroup returning constantly new object for Group causing continuous renders
---
- Replaces Redux with zustand as `initialURL` and `returnToOnAuth` store
- Re-writes React Navigation `getInitialURL` linking function and useOpenInitialURL hook to use new store
- Removes related and some unrelated unused code


thomasgwatson (approved):
lgtm

PR #321, opened 2025-02-28
Create group review

thomasgwatson:
- [x] Test/Fix Create Group
  - [x] Color of Visibility/Accessibility checkboxes is off--barely visible (think changed with recent redesign stuff to a theme color. they should remain white for current design)
  - [x] Presented me with an empty list of member/parent group options when there should have been one
  - [x] Completing creation didn't forward me to the Home of the new Group. Was expecting it would.


apps/mobile/src/components/RoundCheckBox/RoundCheckbox.js
```diff
@@ -53,7 +48,7 @@ const RoundCheckbox = ({
             <Icon
               name={icon}
               size={size * 0.8}
-              color={iconColor}
+              color="hsl(var(--background))"
```
lorenjohnson, line 51: no double quotes

apps/mobile/src/navigation/linking/index.js
```diff
@@ -140,7 +140,7 @@ export const routingConfig = {
   '/:context(groups)/:groupSlug/:streamType(proposals)':                  `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
   '/:context(groups)/:groupSlug/:streamType(requests-and-offers)':        `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
   '/:context(groups)/:groupSlug/:streamType(resources)':                  `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Stream`,
-  '/:context(groups)/:groupSlug':                                         `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab`,
+  '/:context(groups)/:groupSlug':                                         `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Chat Room`,
```
lorenjohnson, line 143: You may already be in this, but I am planning to still zoom out and handle the top level like this. Given that we have a homeWidget / homeURL I wasn't expecting this to be pinned down at the level of linking, but I guess it works for now. We may revisit?

apps/mobile/src/screens/ChatRoomWebView/ChatRoomWebView.js
```diff
@@ -10,7 +10,8 @@ export default function ChatRoomWebView () {
   const navigation = useNavigation()
   const route = useRoute()
   const [{ currentGroup, fetching }] = useCurrentGroup()
-  const { topicName, originalLinkingPath } = useRouteParams()
+  const { topicName: routeTopicName, originalLinkingPath } = useRouteParams()
+  const topicName = routeTopicName || 'home'
```
lorenjohnson, line 14: That looks like it'll solve something nicely 👍

apps/mobile/src/screens/CreateGroupFlow/CreateGroupParentGroups.js
```diff
@@ -37,43 +55,65 @@ export default function CreateGroupParentGroups ({ navigation }) {
...
+          className="flex-row justify-between items-center"
+          onPress={() => groupSelectorModalRef.current.show()}
+        >
+          <Text className="text-foreground/90 font-bold">{t('Parent Groups')}</Text>
+          <View className="border border-foreground rounded-full p-1">
+            <Icon className="text-foreground text-xl" name='Plus' />
+          </View>
+        </TouchableOpacity>
+        
+        <ItemSelectorModal
```
lorenjohnson, line 76: Bravo. Love to see this component in use, and good job figuring it out. Haven't ran this yet, but am wondering if we just want an inline ItemSelector here instead? Will look, and if it works we don't need to do more now, but worth considering.

apps/mobile/src/screens/CreateGroupFlow/CreateGroupReview.js
```diff
@@ -51,86 +51,85 @@ export default function CreateGroupReview () {
...
-            <EditButton onPress={() => navigation.navigate('CreateGroupVisibilityAccessibility')} />
+          <View className="mb-4 border-b border-foreground/20">
+            <View className="flex-row justify-between items-center">
+              <Text className="text-foreground/90 font-bold">{t('Who can see this group?')}</Text>
+              <EditButton onPress={() => navigation.navigate('CreateGroupVisibilityAccessibility')} />
+            </View>
+            <TextInput
+              className="text-foreground text-lg font-bold my-2.5"
+              multiline
+              value={visibilityDescription(groupData.visibility)}
```
lorenjohnson, line 112: Noting for later that this could be resolved on GroupPresenter as `group.visibilityDescription`... When it can be resolved in the presenter I think it is best. If you need the function in Web before conversion then just export it but still called it `visibilityDescriptionResolver`. I don't care a whole lot that you do this right now, but that is my opinion on this for perhaps later work/discussions.

apps/mobile/src/components/Button/ButtonNW.js
lorenjohnson, line 1: Is there any reason we can't just overload Button to support both these cases (and className) and existing cases? I would prefer that. Also, I think `customIconRender` could just be `IconComponent` with a default for `Icon`.

apps/mobile/src/components/RoundCheckBox/RoundCheckbox.js
```diff
@@ -1,17 +1,13 @@
...
 
 const RoundCheckbox = ({
   onValueChange,
   icon = 'checkmark-outline',
   size = 24,
-  checkedColor = caribbeanGreen,
-  uncheckedColor = rhino50,
-  iconColor = 'white',
-  checked = false,
-  style = {}
```
lorenjohnson, line 14: I'd prefer we keep the `style` overrides. I've spend a few hours playing with different scenarios, and I don't see use using NativeWind for everything always and only. IMHO we should be, and will need to, support both `className` and `style`. I care about this in commonly used components (such as this), not so much in full/screen components where it is all domain specific.

lorenjohnson (approved):
Please read through my notes. The only things that I would like to see possible action on is a single cases of double quotes, and my comments on `ButtonNW`.

apps/mobile/src/navigation/linking/index.js
thomasgwatson: Yeah, I was trying to pin down whether context switching was happening from links, so this change is incidental. I'll revert

apps/mobile/src/components/Button/ButtonNW.js
thomasgwatson: I've just had a crack at overloading the one component a few times (actually it was GroupsList that made me very sad), and it lead to some weird inconsistencies that eventually just got me to the "fuck it, just make it primarily nativewind/classnames focused".

So this was done after that. I left them both in place for now; since I didn't want to go on a button-changing rampage but ultimately that is what I think needs to come soon. I've been holding off on doing something like this because its also a lingering issue on web and I'm waiting to see what Aaron does with the web buttons before going ham

apps/mobile/src/navigation/linking/index.js
lorenjohnson: Ah yes, no, it is not. That is a part of the plumbing which remains. I will do that part shortly. That's what I mean by the top end of the linking/routing :)

apps/mobile/src/components/Button/ButtonNW.js
lorenjohnson: I get it. Maybe put a comment to this effect in the header of the component, though I now get the drift and will remember. So basically Button becomes Button again once everything settles and we can are ready for a rampage.

lorenjohnson:
Simulator Screenshot - iPhone 15 Pro Max - 2025-03-01 at 13 02 13
Simulator Screenshot - iPhone 15 Pro Max - 2025-03-01 at 13 02 31
Just pulled this down and doing a design and functional test/review now. 

- I didn't get forwarded to the new group after creating. This was as a new user and my first group. Likely the group wasn't yet fully loaded or reflected in the cache before it was navigated to. Likely we need to either directly re-query at the end of the creation or setup an URQL update cache invalidation.
- Things don't look good/right to in this design, the previous version looked a lot tidier (and I had spent a lot of time on making it so in the past), and I don't yet see how this new design aligns with what is on Web? I am confused and annoyed about this one. 
- Keyboard handling for a group purpose more than a few lines on a smaller device won't work (the multiline TextInput will go below the keyboard without scroll). I just fixed this on GroupWelcome and can fix it here, but something to be aware of. ScrollViews and keyboards are tough on iOS in particular, but I as of the latest work I in the group GroupWelcome area I think we have a good system worked-out that will simply things here. Will apply it. 
- Linting: Double-quotes everywhere. Not sure how I missed this before, but unless we are switching our linting standard can we please just follow StandardJS? Again, happy to get on a call with you and do support if you can't get linting to run.

PR #324, opened 2025-03-01
Drawer experience

thomasgwatson:
Long press to see the names. Even if you scroll after long press, names will remain. Moved context menu styles closer to web

Knocks a few more things off #323 

Screenshot 2025-03-01 at 9 08 11 PM
Screenshot 2025-03-01 at 9 15 01 PM


tibetsprague:
the description says names only appear on long press, but the code makes it look like they appear on scroll too. I think they should appears as soon as you start scrolling too

tibetsprague (approved):
yay! only other comment is I think the name should be a lighter color so easier to read

tibetsprague:
also for groups that extend on to a second line might be better to not let them wrap and show the first some number of letters then a ... then the last number of letters? I mean wrapping one line seems to work but a long one might wrap onto a third line and cause issues

thomasgwatson:
> the description says names only appear on long press, but the code makes it look like they appear on scroll too. I think they should appears as soon as you start scrolling too

Fought back and forth with this to sync up two different triggers and I'm just resetting to a single handler (long-press) for now.

thomasgwatson:
Made the background of the text lighter, to increase the contrast. Maybe a good cue for us to actually set text colors specifically in the themes, instead of intermingling them with the other colors?

simulator_screenshot_D421F73F-9E36-457B-969F-3658AAEA0914

tibetsprague:
we have text colors defined in our theme (all the foreground colors are the text colors), you mean we should set them at a higher level in the component hierarchy?

tibetsprague:
if we have to choose between long press and scroll showing the names, then i would choose scroll. long press is hard to find, but everyone will be scrolling the list and want to see the names then. is there a reason long press is better?

tibetsprague (approved):
I think we probably want to switch to scroll based expansion but happy to start here.

lorenjohnson (approved):
Tom this looked pretty good overall, glad you got to it, and approving here!

I hope not annoyingly, but I did go ahead and take it the next step and clean some stuff up in `ContextSwitchMenu`, and the animation setup as well. I worked through to a pretty good "expand on scroll" + expand on long press behavior which I think is working pretty nicely. Anything I'd have commented on in this PR is reflected in my amendments.

Let's get this merged though as my extension of this work is based on your work in this follow-up PR. Please take a look and I think you'll see a few things I'd want to point out. Mostly I think we can/should apply some NativeWind stuff more selectively, and to keep an eye on putting flex stuff as high up in the component hierarchy as appropriate / possible. I think we should and may need to be a bit more picky about this in Mobile than on Web for performance and layout reasons.  I was able to greatly reduce the amount of styling in my version, so take a look after we get your work here merged. 

https://github.com/Hylozoic/hylo/pull/329

PR #328, opened 2025-03-02
Catches-up mobile linting

lorenjohnson:
After this is merged you can simply run `standard apps/mobile` from terminal in the root of the monorepo and you see now linting issues. You can also run `standard apps/mobile --fix` to auto-fix anything possible to auto-fix. If StandardJS VS Code integration is not working then just run this and fix what it finds as part of completing a PR.


thomasgwatson (approved):
lgtm

PR #329, opened 2025-03-02
Drawer expand on scroll, context menu extend background into safe areas

lorenjohnson:
- Auto-expands ContextSwitchMenu on scroll (and still on long press), currently re-collapses after 2.5 seconds of not interacting. That is adjustable via a constant STAY_EXPANDED_DURATION. 
- Extends backgrounds of ContextSwitchMenu and ContextMenu into safe areas (with padding applied in each component)
- Simplifies animation and styling setup for both menus

https://github.com/user-attachments/assets/47df83b9-261c-460e-b880-b516ca0a45bf


apps/mobile/tailwind.config.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
         background: 'hsl(var(--background) / <alpha-value>)',
         midground: 'hsl(var(--midground) / <alpha-value>)',
         foreground: 'hsl(var(--foreground) / <alpha-value>)',
+        selected: 'hsl(var(--selected) / <alpha-value>)',
```
lorenjohnson, line 26: I'm baffled why these changes show in my branch and in difference to the base branch of `drawer-experience`? I didn't make them, but I do think these are things that were going forward 🤷🏼‍♂️

apps/mobile/tailwind.config.js
lorenjohnson, line 1: I'm baffled why these changes show in my branch and in difference to the base branch of `drawer-experience`? I didn't make them, but I do think these are things that were going forward 🤷🏼‍♂️

thomasgwatson: sometimes I've found that switching the branch its based on doesn't work how I expect and there are residue changes being tracked in the PR

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -11,29 +12,57 @@ import { widgetUrl as makeWidgetUrl } from 'util/navigation'
...
-        setIsExpanded(false)
-      }, 200)
-    }
-    return () => {
-      if (timeoutId) clearTimeout(timeoutId)
-    }
-  }, [isExpanded, isPressing])
+  const collapseTimeout = useRef(null)
+  const pressTimer = useRef(null)
+  const isScrolling = useRef(false)
```
thomasgwatson, line 32: So for these sorts of things, better to useRef than useState? I want to understand when I should be going that path in the future?

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -42,167 +71,109 @@ export default function ContextSwitchMenu ({ isExpanded, setIsExpanded }) {
...
-          </View>
-          {isExpanded && (
-            <Text className='text-xl font-medium text-foreground flex-1 px-2 py-1 bg-foreground/20 rounded-md'>Support</Text>
-          )}
-        </TouchableOpacity>
-      </View>
-    </Animated.View>
+      <ContextRow
+        bottomItem
+        isExpanded={isExpanded}
```
thomasgwatson, line 100: for fwiw, I think these two should stay the same, regardless of isExpanded

tibetsprague:
Noting that Dev on web now has a version of this that Aaron designed. Honestly I don’t love it, so we’ll probably iterate on that more too. Let’s see if we can get something that feels somewhat similar on web and app?On Mar 3, 2025, at 6:13 AM, Tom Watson ***@***.***> wrote:﻿
@thomasgwatson commented on this pull request.



In apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js:
> -          onPressOut={handlePressOut}
-          delayLongPress={200}
-          className='flex flex-row items-center gap-3 justify-start p-1'
-        >
-          <View className='bg-primary relative flex flex-col items-center justify-center w-14 h-14 min-h-10 rounded-lg drop-shadow-md opacity-60 scale-90'>
-            <CircleHelp />
-          </View>
-          {isExpanded && (
-            <Text className='text-xl font-medium text-foreground flex-1 px-2 py-1 bg-foreground/20 rounded-md'>Support</Text>
-          )}
-        </TouchableOpacity>
-      </View>
-    </Animated.View>
+      <ContextRow
+        bottomItem
+        isExpanded={isExpanded}

fwiw, I think these two should stay the same, regardless of isExpanded

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
lorenjohnson: Not following... `isExpanded` is currently used here to reveal or hide the title, and that behavior is the same for both these bottom items and the contexts.

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
lorenjohnson: Well I had to think for a moment about the real reason, but the official answer is first off to remember/understand that useState is basically a useRef + triggering re-renders on every "set*" call. So in the case of story the timeout and flags, using useRef means that we save some re-renders as otherwise every time any of these values were "set" the component would re-render. But these particular variables are used to derive states that already make re-renders (i.e. `isExpanded` / `setIsExpanded`). In more practical everyday use I end-up with useRef to reduce complexity as I can deal with things in the same render cycle more reliably. I am getting the current value in the current render state with myRef.current vs when I useState and I am only getting that update after the next render. That all said, I usually only get around to thinking about useRef when it is recommended to me via a StackOverflow code sample or ChatGPT and such. In this case it was a back and forth with the GPT that ended-up here. 

_Note that in current code on dev actually we only have collapseTimeout and none of these other variables/refs, as I was able to further simplify things here in a covert quick refactor that is now on dev. Take a look and you'll see what I mean._

PR #346, opened 2025-03-04
fix bug: question marks in proposal options

### LINKED ISSUE #345: ? - Question marks not being displayed correctly in Proposals
When creating a proposal '?' are displayed incorrectly after the proposal is posted.

See last option of multiple choice in attached screenshot.

I edited it twice to double check - the bug remains. I checked a few other post types, '?" were displayed correctly elsewhere.

Using Safari macOS 15.3

https://hylo.canny.io/admin/board/bugs-feature-requests/p/question-marks-not-being-displayed-correctly-in-proposals

KevinTriplett:
closes #345 

Due to SQL injection protection, "?" was replaced with "$n", where n was a sequential number.

I recommend we audit the code to find all instances of where we may be allowing SQL injection.

### DIFF
--- a/apps/backend/api/models/Post.js
+++ b/apps/backend/api/models/Post.js
@@ -419,11 +419,7 @@ module.exports = bookshelf.Model.extend(Object.assign({
 
     // Execute the insert query for options passed in
     if (options.length > 0) {
-      const insertQuery = `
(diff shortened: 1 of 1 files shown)


thomasgwatson (approved):
lgtm

PR #347, opened 2025-03-04
Move notifications from TabStackHeader to TabsNavigator

thomasgwatson:


apps/mobile/src/navigation/TabsNavigator.js
```diff
@@ -2,18 +2,23 @@ import React from 'react'
...
 import useCurrentUser from '@hylo/hooks/useCurrentUser'
+import { modalScreenName } from 'hooks/useIsModalScreen'
 import Icon from 'components/Icon'
 import {
   alabaster, black10OnCaribbeanGreen, gainsboro, gunsmoke, white
 } from 'style/colors'
 import HomeNavigator from 'navigation/HomeNavigator'
 import SearchNavigator from 'navigation/SearchNavigator'
 import MessagesNavigator from 'navigation/MessagesNavigator'
+import { useNavigation } from '@react-navigation/native'
```
lorenjohnson, line 15: FYI - you probably have inferred by now that I have a preference for how import statements are ordered. My preference is basically as follows. I don't need in any action here, but you'll likely see me put shuffle things into this order whenever I touch a file that isn't already such:

1.	External Libraries – Core dependencies first (react, react-native), followed by utilities (lodash), data/state (redux, urql), and navigation.
2.	Project Files – Utilities, hooks, API queries/mutations, state management, and business logic. `@hylo` libraries at the end of this list. 
3.	Components & Styles – Components always come almost last, ideally imported in alphabetical order
4. `styles` / colors as the final imports if they exist.

apps/mobile/src/navigation/TabsNavigator.js
```diff
@@ -52,6 +57,20 @@ export default function TabsNavigator () {
     <Tabs.Navigator {...navigatorProps}>
       <Tabs.Screen name='Home Tab' component={HomeNavigator} />
       <Tabs.Screen name='Messages Tab' component={MessagesNavigator} options={{ tabBarBadge: messagesBadgeCount }} />
+      <Tabs.Screen
+        name='Notifications Tab'
+        component={HomeNavigator} // it will never navigate to this but we need to pass a valid component here anyway
```
lorenjohnson, line 62: You can just `() => {}` this (or create above outside the TabsNavigator component a DummyComponent with that definition)  and remove the comment. I think that would be clear enough and also not ever risk something rendering here that shouldn't . if it breaks it would be more clear.

lorenjohnson (approved):
LGTM. Would prefer you do something with the two things I mention, but your call.

lorenjohnson:
@thomasgwatson packaging up a Bitrise build with all current stuff and doing a round of zoomed-out testing, and want this in there. I made the couple linting updates I mentioned, and also decided to move the Notifications icon to the far right on the bottom Tab bar instead of having Search there. Maybe that is wrong, and if so happy to reverse it. As it is merging like this as I think you may be out for a couple days and want to keep moving.

thomasgwatson:
Sweet

PR #348, opened 2025-03-04
DRAFT: Open post details and member views in modals at the root navigation l…

thomasgwatson:
…evel

A few follow up things touched on here:
https://github.com/Hylozoic/hylo/discussions/323#discussioncomment-12384656

TODO:

- [x] switch burger to back chervon


lorenjohnson (commented):
@thomasgwatson I’ve reviewed and tested what is here, and it is was informative to see what it'd feel like to just have Post Detail open in a 2nd level Modal always. It didn't feel as off as I thought it might. There is currently an app-crashing bug on nav from the Stream to a Member because of a missing import in PostCard, and there are a variety of cases where I can still end-up on a non-modal stack screen which aren't yet covered here. I think we should go a different direction, and not change the architecture of where these screens are located, how we navigate to them, or how they are named. I would instead introduce a "modal-like" header with an X instead of a back button to those screens while leaving them in the stack and not in the "2nd level" modal pop-out style).  I think. If we are going to only use the AuthRoot level modals then I want "full screen" style modal for these areas vs the pop-out modal so the distinction between what is "in context" and not is maintained. So please hold-off on merging this or going further in the particular implementation for the moment. I think I may just take this one over, if you're up to directing to the next/another task?

thomasgwatson:
Yep yep.

I'll work through the Welcome Page changes

PR #354, opened 2025-03-04
chore(web,mobile): Fixes HyloEditorMobile but probably breaks it for Web

lorenjohnson:
Changing `onCreate(editor)` to `onCreate({ editor })` gets the content loading again, but changes don’t save without also removing the `onFocus`, `onBlur`, and initialized stuff. Also, this change of signature for `onCreate` may reflect a change made in Web code, so something may need to be updated in Web to go back to this function signature? These callback methods I’ve tried to keep with the same signature of the underlying useEditor/TipTap callbacks for simplicity. Don’t want to commit this as it is gonna break something in Web I’m sure, but here this branch/PR does work on Mobile for reference, and maybe we can work back from here?

### DIFF
--- a/apps/web/src/components/HyloEditor/HyloEditor.js
+++ b/apps/web/src/components/HyloEditor/HyloEditor.js
@@ -29,8 +29,6 @@ const HyloEditor = React.forwardRef(({
   onEnter,
   onAltEnter,
   onEscape,
-  onFocus,
-  onBlur,
   placeholder,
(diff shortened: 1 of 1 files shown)


tibetsprague:
this seems to all work fine on web. I should have tested and tried to understand what Aaron was doing more. I dont think any of it is used/needed

tibetsprague (approved):

PR #361, opened 2025-03-06
334 scroll the menu to new widget 2

### LINKED ISSUE #334: Adding new widget to menu should scroll the menu to that widget
When editing the context menu. Add View. Create a new chat room. Select a topic. Hit create. The new chat room widget has been added to the menu at the bottom. If the menu is long you won’t see where it was added so you may not realize it had been added. After adding the new widget, we should scroll the context menu to the new widget so it is visible.

### LINKED ISSUE #360: Child widgets fly back to original spot when moving
Add a container. Add two views to that container. When dragging (moving) a child widget and dropping, the child widget animates a "fly back" to its original location but then is rendered in the new location.

KevinTriplett:
closes #334
closes #360 

This is a new branch that solves the above referenced issues plus adjacent issues generously discovered by @brodeur.

* clicking "Edit Menu": Context Menu stays at bottom to prevent UX disorientation
* add a widget: context menu scrolls to the widget and pulsates a glow to draw attention
* add a child view: no scrolling since it appears above the add view button
* move a view or widget:
  * prevent context menu from scrolling back to original view / widget position
  * constrain to vertical
* make `<ul>` html compliant by wrapping all elements in `<li>` tags


KevinTriplett:
Oooh ... I found out that when moving a child view to another container, the CM scrolls back to where the child view originally way, which is waay confusing. I'll try to fix that now.

KevinTriplett:
Ah ... it's because when the user drops the widget, it flies back to its original location, so CM scrolls back to the original location. I've filed a new issue #360 and that should fix this issue of moving child views to "far away" drop spots.

KevinTriplett:
In the above commits, the first closes #360 and the second constrains the drag to the vertical axis.

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -84,6 +92,37 @@ export default function ContextMenu (props) {
   const [activeWidget, setActiveWidget] = useState(null)
   const toggleNavMenuAction = useCallback(() => dispatch(toggleNavMenu()), [])
 
+  const currentWidgetIds = widgetIdsToArray(orderedWidgets)
+  const newWidgetId = previousWidgetIds.length > 0 ? currentWidgetIds.find(widgetId => previousWidgetIds.indexOf(widgetId) === -1) : null
```
tibetsprague, line 96: i would put this in a useMemo so its only calculated when previousWidgetIds or currentWidgetIds changes

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -84,6 +92,37 @@ export default function ContextMenu (props) {
...
+  previousWidgetIds = currentWidgetIds
```
tibetsprague, line 97: i think this should be a state variable that updates in a useEffect when currentWidgetIds changes

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -35,6 +36,13 @@ import getQuerystringParam from 'store/selectors/getQuerystringParam'
 import logout from 'store/actions/logout'
 import classes from './ContextMenu.module.scss'
 
+let previousWidgetIds = []
+let isAddingChildWidget = false
```
tibetsprague, line 40: this looks like it should be a state variable

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -185,22 +226,26 @@ export default function ContextMenu (props) {
   )
 }
 
-function ContextWidgetList ({ contextWidgets, groupSlug, rootPath, canAdminister, isEditing, isDragging, activeWidget, group }) {
+function ContextWidgetList ({ contextWidgets, groupSlug, rootPath, canAdminister, isEditing, isDragging, activeWidget, newWidgetId, newWidgetRef, group }) {
   const navigate = useNavigate()
   const location = useLocation()
 
   const handlePositionedAdd = ({ widget, addToEnd, parentId }) => {
+    isAddingChildWidget = true
```
tibetsprague, line 234: is this variable even in this scope?

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -185,22 +226,26 @@ export default function ContextMenu (props) {
...
     navigate(addQuerystringToPath(location.pathname, { addview: 'yes', cme: 'yes', parentId: widget?.parentId || parentId, orderInFrontOfWidgetId: widget?.id || null }))
   }
 
+  const itemProps = {}
+  if (newWidgetId) itemProps[newWidgetId] = { ref: newWidgetRef }
```
tibetsprague, line 239: why is itemProps having a key of the newWidgetId?

tibetsprague (commented):
the functionality works great!

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -35,6 +36,13 @@ import getQuerystringParam from 'store/selectors/getQuerystringParam'
 import logout from 'store/actions/logout'
 import classes from './ContextMenu.module.scss'
 
+let previousWidgetIds = []
+let isAddingChildWidget = false
+
+const widgetIdsToArray = (widgets) => {
```
thomasgwatson, line 42: You could probably strip this function out and just run map(widget => widget.id) on the rawContextWidgets inside a useMemo. rawContextWidgets will be a flat list of a group's context widgets

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: Unless you see how to accomplish these simply and we're avoiding module variables at all costs, I'd like to keep these here as module variables for these reasons:
* previousWidgetsIds needs to persist between renders
* isAddingChildWIdget is referenced by two different module functions
* a state variable would trigger a re-render

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: Yes, as a module variable (see above comment)

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: To create a ref to the correct widget -- lint wouldn't allow a `const` into the `contextWidgets.map` function, and react wouldn't let me put a conditional statement to the `li` tag. This turned out to be the first acceptable mechanism to conditionally assign a ref inside an iterable that worked. So I stopped there.

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: I added a useMemo to set newWidget only when `currentWidgetIds` changes

tibetsprague (approved):
I suspect there are slightly more efficient ways to handle the state but its working and this is good enough for now

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: DONE - works

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: The problem with state variables is that they trigger a re-render -- useMemo works well here

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: I started with rawContextWIdgets but the functionality was working. According to `orderContextWidgetsForContextMenu`, this may provide a clue:

`// Step 1: filter out widgets without an order, as these are not displayed in the context menu`

KevinTriplett: I did move that function into the useMemo function, so it's no longer a module function 👍🏼

PR #362, opened 2025-03-06
WIP switch to React Router data router

tibetsprague:


thomasgwatson:
Assuming this is so out of date its not worth keeping around

PR #367, opened 2025-03-06
Context switching links, Improved UX, openURL updates, etc

lorenjohnson:
- Properly handles context switching links and adds code in HomeNavigator re. flow of initialURL, returnToOnAuth, and context switching link handling (as it all happens in HomeNavigator)
- Makes TabNavigationHeader (the header almost always seen in Mobile) have a < left button always
- Gives mostly full width of header to title of group
- Keep title of group in header even on custom stream types, and moves the custom stream type name (e.g. Events, Proposals, etc) in the stream header
- Makes stream header much smaller and adds a "+ Create" button instead of the Post Prompt
- Fixes some things in Stream
- Extends openURL to have "options" as the second param, with `openURL(path, { reset: true})` now being the signature for what used to be `openURL(path, true)`. Adds options.replace option which should be what is generally always used when using openURL to navigate from either Drawer menu (ContextSwith or Context Menu) so that the Stack on the Home tab is reset by what is navigated to vs creating addiotnal screens/history in the stack. _Replace should probably not be used other than from this top level, as we like the history to be correct. Reset should never be used with the exception of the JoinGroup / Auth related places it is currently used._
- Updates ContextSwitch and ContextMenu (including widget paths) to navigate using replace as per above
- Gives TabNavigationHeader a white background and adjusts font and icon sizes somewhat
- Fixes issue with Map back navigation
- Minor UX improvement to Creation dialog to make it a bottom sheet

### DIFF
--- a/apps/mobile/src/components/ContextMenu/ContextMenu.js
+++ b/apps/mobile/src/components/ContextMenu/ContextMenu.js
@@ -9,7 +9,7 @@ import useContextWidgetChildren from '@hylo/hooks/useContextWidgetChildren'
 import useHasResponsibility, { RESP_ADD_MEMBERS, RESP_ADMINISTRATION } from '@hylo/hooks/useHasResponsibility'
 import { widgetUrl as makeWidgetUrl, groupUrl } from 'util/navigation'
 import useLogout from 'hooks/useLogout'
-import { openURL } from 'hooks/useOpenURL'
+import useOpenURL from 'hooks/useOpenURL'
 import useRouteParams from 'hooks/useRouteParams'
 import GroupMenuHeader from 'components/GroupMenuHeader'
 import WidgetIconResolver from 'components/WidgetIconResolver'
@@ -18,6 +18,7 @@ export default function ContextMenu () {
   const insets = useSafeAreaInsets()
   const navigation = useNavigation()
   const { t } = useTranslation()
+  const openURL = useOpenURL()
   const [{ currentGroup, fetching }] = useCurrentGroup()
   const widgets = useMemo(() =>
     orderContextWidgetsForContextMenu(currentGroup?.contextWidgets || []),
@@ -25,7 +26,7 @@ export default function ContextMenu () {
 
   useEffect(() => {
     if ((!fetching && currentGroup?.shouldWelcome)) {
-      navigation.navigate('Group Welcome')
+      navigation.replace('Group Welcome')
     }
   }, [fetching, currentGroup])
 
@@ -37,7 +38,8 @@ export default function ContextMenu () {
       },
       rootPath: `/groups/${currentGroup?.slug}`,
       groupSlug: currentGroup?.slug
-    })
+    }),
+    { replace: true }
   )
 
   if (!currentGroup) return null
@@ -75,6 +77,7 @@ export default function ContextMenu () {
 function MenuItem ({ widget, groupSlug, rootPath, group }) {
   const { t } = useTranslation()
   const { listItems, loading } = useContextWidgetChildren({ widget, groupSlug })
+  const openURL = useOpenURL()
   const logout = useLogout()
   const hasResponsibility = useHasResponsibility({ forCurrentGroup: true, forCurrentUser: true })
   const canAdmin = hasResponsibility(RESP_ADMINISTRATION)
@@ -93,7 +96,7 @@ function MenuItem ({ widget, groupSlug, rootPath, group }) {
 
   const handleWidgetPress = widget => {
     const linkingPath = makeWidgetUrl({ widget, rootPath, groupSlug: currentGroup?.slug })
-    openURL(linkingPath)
+    openURL(linkingPath, { replace: true })
   }
 
   if (isHiddenInContextMenuResolver(widget)) return null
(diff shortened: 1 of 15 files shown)


lorenjohnson:
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-06 at 18 57 47
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-06 at 18 58 21

apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -18,14 +18,15 @@ export default function ContextMenu () {
   const insets = useSafeAreaInsets()
   const navigation = useNavigation()
   const { t } = useTranslation()
+  const openURL = useOpenURL()
```
thomasgwatson, line 21: is this how we should be using openURL from here on out?

lorenjohnson: Yes, always preferable, as the non-hook version relies upon a global navigationRef vs getting it through context. The difference is academic for now, but the global ref could become unstable. So yes, but this is just clean-up though, I wouldn't fret much about it.

lorenjohnson:
Also, make an update to Creation modal to make it a bottom sheet. Totally just a bonus thing, but gives us a place to reference for any other bottom sheet style things we may wish to do elsewhere and a minor UX improvement: Simulator Screenshot - iPhone 16 Pro Max - 2025-03-07 at 00 27 55

apps/mobile/src/navigation/linking/index.js
```diff
@@ -59,7 +59,8 @@ export const routingConfig = {
   // This route isn't correct; no such screen exists. And its not to be confused with the Group Explore screen.
   '/:context(public)/groups':                                             `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Group Explorer`,
   '/:context(public)/map':                                                `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Map`,
-  '/:context(public)/:groupSlug/post/:postId':                            `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Post Detail`,
+  // TODO: Is this a valid route? / Ever used?
+  '/:context(public)/:groupSlug/post/:id':                                `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Post Details`,
```
tibetsprague, line 63: not right now, but I could imagine it

tibetsprague (approved):
lgtm

apps/mobile/src/screens/Stream/StreamHeader.js
```diff
@@ -47,53 +45,23 @@ export function PostPrompt ({ forGroup, currentType, currentTopicName }) {
...
   return (
-    <View style={styles.postPrompt}>
-      <TouchableOpacity onPress={handleOpenPostEditor} style={styles.promptButton}>
-        <Avatar avatarUrl={currentUser.avatarUrl} style={styles.avatar} />
-        <Text style={styles.promptText}>{postPromptString(currentType, currentUser, t)}</Text>
-      </TouchableOpacity>
-    </View>
+    <TouchableOpacity className='flex-row rounded-lg p-2 bg-primary opacity-90' style={styles.postPrompt} onPress={handleOpenPostEditor}>
+      <Plus style={{ color: rhino }} size={20} />
+      <Text style={{ marginRight: 5 }} className='text-l'>Create</Text>
```
thomasgwatson, line 50: Need to i18n this Create here; I'll get to it if you don't

apps/mobile/src/navigation/AuthRootNavigator.js
```diff
@@ -153,32 +153,31 @@ export default function AuthRootNavigator () {
...
         <AuthRoot.Screen name='Drawer' component={DrawerNavigator} options={{ headerShown: false }} />
-        <AuthRoot.Screen
-          name='Create Group' component={CreateGroupTabsNavigator}
-          options={{ headerShown: false }}
-        />
+        <AuthRoot.Screen name='Create Group' component={CreateGroupTabsNavigator} options={{ headerShown: false }} />
+        <AuthRoot.Screen name='Loading' component={LoadingScreen} options={{ headerShown: false, animationEnabled: false }} />
+        {/*
+          == Modals ==
+          modelScreenName is used to differentiated screen names from ones that have a non-model counterpart
```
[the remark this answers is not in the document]
lorenjohnson: That was not clear at all before, and as it is this '- Modal' pattern and the accompany hook is used internally in components to recognized when rendered within a modal to do different things accordingly. For now don't use that pattern unless it is needed for something which is mostly evidenced by finding yourself naming a screen something like "*modal/Modal". I'd just make all modals use modalScreenName but actually would like to do away with the pattern over time vs extend, if that makes sense.

apps/mobile/src/navigation/linking/index.js
lorenjohnson: Not sure where it crept in, but will remove it. I never keep code that might be used.

thomasgwatson (approved):
lgtm

PR #368, opened 2025-03-06
Fix case where when socket not connected you could create a chat post…

tibetsprague:
… and then clicking on it would break the UI


thomasgwatson:
Is this one ready to go? I would have guessed there would be more changes on web to make this happen?

tibetsprague:
yup, ready to go! the changes on web already happened, this is just one last part to it that I didnt think was necessary previously

thomasgwatson (approved):
lgtm

PR #374, opened 2025-03-07
Batch of Mobile refactors and UX improvements

lorenjohnson:
- Another round of re-factor for CreateGroup as it seems to have been prototypical to the GroupWelcome workflow and it was a pretty old/unnecessarily brittle and complicated implementation unnecessarily utilizing TabsNavigator. There were also a number of subtle and not-so-subtle UX issues there still which should be now cleaned-up (e.g. back/continue buttons appeared below keyboard, single textinput entries weren't set to auto-focus, etc). @thomasgwatson if you haven't already done the upcoming GroupWelcome update please reference this current implementation for that if possible. 
- Consolidates all the initialURL, returnToOnAuth, and current group/context switching via links into a single `useHandleLinking` hook which is now promoted up from HomeNavigator to AuthRootNavigator as there are cases where it existing in HomeNavigator were potentially missing some incoming links. This new setup has been tested, but still may need to shuffle around more as we test linking. Overall though it is easy to move around maintain in this form. 
- Minor update to `ButtonNW` such that its default style is not dependent on static sizing (always to be avoided when possible), and just looks ok out-of-the-box. Works well in Create Group, but haven't looked at how it looks in GroupWelcome yet, which was the only other place this new Button was used, but I imagine it won't change much there if not just make the buttons look a little better.
- Adds an additional option to changeToGroup to optionally skip the check of whether the user can view the group. This check is unreliable in some cases, like when adding a new group, but is good to keep still as the default. The `confirm` option which was default of `true` is now defaulted to false, and is one of the options values vs simply the second parameter. The usage of changeToGroup across the app was changed to reflect this update appropriately.
- Removes the extra 2nd-level header/backbutton from GroupSettings, the React Navigation top-level back button is now setup in that component to handle that.


apps/mobile/src/screens/CreateGroup/CreateGroup.js
lorenjohnson, line 1: This plus a bit of state management in the associated zustand store replaces the previous CreateGroupTabsNavigator setup.

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
lorenjohnson, line 1: Worth noting is that creating a group with a parent group currently seems to be broken, which is not new in this refactor. The issues appears to be backend and is something I'll add a todo or ticket to explore shortly.

apps/mobile/src/screens/GroupSettingsWebView/GroupSettingsWebView.js
```diff
@@ -16,6 +17,12 @@ export default function GroupSettingsWebView ({ path: pathProp, route }) {
     setSelectedSettingsArea(routeSettingsArea)
   }, [routeSettingsArea])
 
+  useEffect(() => {
+    navigation.setOptions({
+      headerLeftOnPress: () => selectedSettingsArea ? setSelectedSettingsArea() : undefined
```
lorenjohnson, line 22: BackButton management for GroupSettings...

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
thomasgwatson: Well, if we actually want to pull this closer to the new create group flow on web, we would actually strip several parts out. Adding a parent group, and adding the purpose would go, and be 'handled' after creation.

thomasgwatson (approved):
lgtm

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
tibetsprague: lets leave in the parent group thing, I think we need to bring it back on web actually (I also think purpose is fine to leave in for now)

PR #376, opened 2025-03-08
Removed pinned post functionality from web UI for now, until we get it working on…

tibetsprague:
… a per view basis

Leave in the back-end for now, for mobile app, and because we may re-use

### DIFF
--- a/apps/web/src/components/PostCard/PostHeader/PostHeader.connector.js
+++ b/apps/web/src/components/PostCard/PostHeader/PostHeader.connector.js
@@ -8,7 +8,6 @@ import removePost from 'store/actions/removePost'
 import {
   unfulfillPost,
   fulfillPost,
-  pinPost,
   getGroup,
   updateProposalOutcome
 } from './PostHeader.store'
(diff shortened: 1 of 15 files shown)


tibetsprague:
@thomasgwatson not the highest priority, but we can remove any pinning UI from the app if it exists

lorenjohnson:
The app does have full pinned post functionality, and it works well. I hear it's coming back, but rather sad to see it go as I have found it quite useful. Was it considered to be just disabled for everything but the Stream common view instead?

tibetsprague:
We could I suppose leave it on the stream but that seems a bit odd and confusing, and we’d have to only show the pin action on posts when viewing them from the stream for it to make sense. Or change the text everywhere to say Pin to Stream. I’ll bring it up again in our product meeting tomorrow, but for now we think its much better to just add a post to the group menu which makes it more findable and also not take up a lot of space at the top of the stream every time you open it. And we will bring back pinning soon where you can pin a post to any view and we will display it in a new smaller UI at the top of the view.

lorenjohnson:
Ah, I see so basically the idea is that the Pinned post functionality is replaced by this other feature/s. That makes sense to me 👍

lorenjohnson:
We could just make the Pin post menu add it to the Group posts... but I will digress for now because I'm really not familiar with the related features yet to say.

lorenjohnson (approved):
LGTM 👍 

_I think in an ideal world we'd do an assisted migration or something for current communities that may be relying on this functionality, but all things considered I think we just move ahead removing it here and on Mobile and maybe if there is a Newsletter or anything going out maybe have a bit in there about the alternative ways to achieve similar functionality. Also it might be good to commit to this and remove from the Backend after Mobile is removed. If this functionality comes back I don't think it will be bad to have to carefully re-implement it with new stream specific requirements and different enough to maybe be better to start from the top._

tibetsprague:
we are definitely going to be giving supportive instructions to group leaders about all the changes

PR #386, opened 2025-03-11
Handle external links for custom views in mobile

thomasgwatson:
external links mainly just need the right protocol applied.

I've done this outside of the nav utilities in this PR but they could also be done inside widgetUrl itself


lorenjohnson (commented):
Surprised to see all external link URL protocol management here. The default case for openURL should just work. Why are we doing http vs https checking? Shouldn't this be up to user or otherwise already sanitized upstream? It was already implemented for custom views in the old
GroupNavigation and don't remember needing to do any special handling.

thomasgwatson:
twicefire.com is a valid external link for a custom view. openURL treats that as an internal path unless the protocol is added at the start.

On web we are doing:

```
  if (externalLink) {
    const url = externalLink.startsWith('http://') || externalLink.startsWith('https://')
      ? externalLink
      : `https://${externalLink}`
```

Could probably just do that as well

thomasgwatson:
So in short, no, its not sanitized upstream. Do you want to do that in makeWidgetURL? Or even elsewhere?

lorenjohnson:
Got it. I of course wish this was handled on the backend either at time of mutation or query, but absent of that we do have a @hylo/shared method for handling this at `TextHelpers.sanitizeURL()`. We use that for preparing projectManagementLink and donationsLink URLs in the PostEditor. I would either use that directly as `openURL(TextHelpers.sanitizeURL(view.externalLink))` to limit the bleed and not build anything up here that we ultimately want to have go away, or just do this upstream in the backend resolver CustomView#externalLink resolver using the same @hylo/shared method, which I do think is the right answer.  This method can be re-ran on the same URL harmlessly so it wouldn't matter if you removed the conversion on Web, though ideally for clarity that would be removed at the same time.

apps/mobile/src/components/ContextMenu/ContextMenu.js
```diff
@@ -96,7 +96,15 @@ function MenuItem ({ widget, groupSlug, rootPath, group }) {
   }, [url, routeParams.originalLinkingPath])
 
   const handleWidgetPress = widget => {
+    if (widget?.customView?.externalLink) {
+      try {
+        openURL(ensureHttpsForPath(widget.customView.externalLink))
+      } catch (error) {
```
lorenjohnson, line 102: If we want to handle the error and just have it console.log then I'd add that to openURL. I think I'd prefer to just not handle this error though so it passes through to Sentry error capture.

apps/mobile/src/util/navigation.js
```diff
@@ -213,10 +212,6 @@ export function removeGroupFromUrl (url) {
   return url.replace(new RegExp(matchForReplaceRegex), '')
 }
 
-export function gotoExternalUrl (url) {
```
lorenjohnson, line 216: This is still used in MemberProfile. It should probably be replaced by openURL there, but if you don't want to convert and test that area then just leave this for now.

lorenjohnson: Strike that, I was wrong. Sorry. Saw it in Web and thought I was seeing a Mobile search result. Yes, removing this is great :)

apps/mobile/src/util/navigation.js
```diff
@@ -225,3 +220,20 @@ export const origin = () =>
...
+export function ensureHttpsForPath (path) {
+  // Check if path is a valid URI path
+  // Regex matches paths that start with a domain name pattern
+  // e.g. example.com, sub.example.com, example.co.uk etc
+  const validPathRegex = /^[a-zA-Z0-9][a-zA-Z0-9-]*(\.[a-zA-Z0-9][a-zA-Z0-9-]*)*\.[a-zA-Z]{2,}(\/.*)?$/
+
+  if (!path || typeof path !== 'string') return null
+
+  // Remove any existing protocol
+  const cleanPath = path.replace(/^https?:\/\//, '')
```
lorenjohnson, line 233: I don't think we should force https. If the user enters an http as protocol it may very well be on purpose, and if not then they can update the URL as they discover it.

lorenjohnson:
I am sorry to be this way, but I really think this is the right solution and fairly ease to test to confirm, which I've done on both Mobile and Web. Really wanting to squash these sort of weirdnesses as we find them: https://github.com/Hylozoic/hylo/pull/390

lorenjohnson:
Resolved in #390

PR #390, opened 2025-03-12
chore(backend,web,mobile): Sanitize CustomView#externalLink on query

lorenjohnson:
_We also might just want to for now abandon relying on the navigation helpers in both apps for ContextWidgets links for now (until that area is sorted and shared) and just put the `widgetUrl` helper in `ContextWidgetPresenter` and add a `presentedWidget#link({ groupSlug, rootPath, ... })` attribute on the presented Widget... Won't do that here, but that would simplify the Web and Mobile code and guarantee we were always getting the same path in both apps for the same `ContextWidget`._

### DIFF
--- a/apps/backend/api/graphql/makeModels.js
+++ b/apps/backend/api/graphql/makeModels.js
@@ -1,5 +1,6 @@
 import { camelCase, mapKeys, startCase } from 'lodash/fp'
 import pluralize from 'pluralize'
+import { TextHelpers } from '@hylo/shared'
 import searchQuerySet from './searchQuerySet'
 import {
   commentFilter,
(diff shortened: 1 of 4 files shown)


thomasgwatson (approved):
lgtm.

With regard to moving widgetURL;
In so far as it means we move all the other required functions, I guess that's fine. It just depends on a bunch of other stuff, so it might defacto be like creating a created navigation util anyway

lorenjohnson:
Yes, I get that. Just noting for soon, but sometime after release, and yes I imagine it will be a changeset that comes with a unified cross-platform set of path generators. With a bit of ChatGPT, care, and testing it won't be a biggy but .... for later.

PR #392, opened 2025-03-12
Redo all our emails!

tibetsprague:
New email templates for every email!

Lots of changes to email digest, including:
- Show posts that are upcoming in the next 2 days (for daily digest) or next week (for weekly digest
- Also show posts that are ending in that timeframe
- Filter out content created by the user being sent to so they dont see their own content
- Filter out chats that have already been seen

Don't auto log people in when they click on any link in their notification emails. Too dangerous.


tibetsprague:
would like a look at this, but going to merge to get testing on staging

PR #395, opened 2025-03-13
ContextMenu updates, Mobile app initial load state

### LINKED ISSUE #403: When creating a post, clicking on the post type doesnt pop up the changer interface
You have to carefully click the down arrow. which is hard to know/find
on ios

lorenjohnson:
- Get initial app load to always do the correct thing given a default of My context or lastViewedGroup if user has memberships, test/fix for incoming context-switching links, etc... Basically initial nav state in app
- Aligns ContextMenu isActive items with current Web style
- Integrates All Views in ContextMenu to use ContextWidget component, now responds to isActive status
- Adds Membes list to ContextMenu
- WidgetIconResolver now resolves to LucideIcon if one found in the current LucideIcon set in Mobile
- Adds allViewsWidget export to ContextWidgetPresenter and applies in both Web and Mobile ContextMenus (web+mobile)
- Removes custom Logout widget handling in render as with updates it renders as expected, calling of logout() on press is intercepted for widget.type === login within handleWidgetOnPress
- Iterates some GroupPresenter naming and methods for more clarity
- Refines / simplifies ContextMenu styling (code-level, mostly not visible differences in UI)
- Adds key to ImageAttachments component iteration (fixes on one of the very long warnings that make debugging via logging so annoying)
- Closes #403, #404, and #406
- Fixes GroupWelcome check redirect with query updater
- Code-level organization in presenters and linting / clarifying throughout
- Adds getCurrentRoute helper for dealing with React Navigation nested state


packages/graphql/fragments/groupFieldsFragment.js
```diff
@@ -91,7 +91,7 @@ export default gql`
       neighborhood
       region
     }
-    members(first: 8, sortBy: "name", order: "desc") {
+    members(first: 4, sortBy: "last_active_at", order: "asc") {
```
tibetsprague, line 94: this should be desc, so its showing most recently active. about to do this on web too

tibetsprague (approved):
lgtm besides one small thing

packages/graphql/fragments/groupFieldsFragment.js
lorenjohnson: will update, or you could update here then just use this fragment ;)

PR #398, opened 2025-03-13
Mobile welcome page

thomasgwatson:
OK: so the overall GroupWelcome redirect will also pick up on the needs of the GroupWelcomePage...
It will also now DISPLAY the content of that page in the prior GroupWelcome...
And at the end of the GroupWelcome, it will update the membership to ensure that lastViewedAt is updated, and that the same user won't trigger the redirect again.

HOWEVER, as discussed with Loren, the mutation updateMembership is not updating the local copy, and so Loren is trying to balance the updating needs of the GroupWelcome and the flicker on the Stream (which also calls this mutation). So this feature is dependent on those changes to work properly


packages/graphql/fragments/groupPrerequisiteGroupsFieldsFragment.js
```diff
@@ -16,6 +16,7 @@ export default gql`
           locationDisplayPrecision
           publicMemberDirectory
           showSuggestedSkills
+          showWelcomePage
```
lorenjohnson, line 19: Does this status happen to reflect at all on the status for the existing Group Welcome? Probably not, but the current function in the presenter for getting that status is hairy, and I sure would love it if there was a backend thing just gave me a boolean as to whether we should prompt for agreements or not:

```
const getShouldWelcomeResolver = group => {
  return currentUser => {
    if (!group || !currentUser) return false
    if (isStaticContext(group.slug)) return false

    const currentMembership = currentUser?.memberships &&
      currentUser.memberships.find(m => m.group.id === group?.id)

    const { agreementsAcceptedAt, joinQuestionsAnsweredAt, showJoinForm } = currentMembership?.settings || {}

    const numAgreements = group?.agreements?.total || 0

    const agreementsChanged = (!isStaticContext(group?.slug) && numAgreements > 0) &&
      (!agreementsAcceptedAt || agreementsAcceptedAt < group?.settings?.agreementsLastUpdatedAt)

    return (
      (!isStaticContext(group?.slug) && showJoinForm) ||
      agreementsChanged ||
      (group?.settings?.askJoinQuestions && !joinQuestionsAnsweredAt)
    )
  }
}
```

thomasgwatson: So, as you'll see, I added another line to the logic party down the bottom of this. Agreed its hairy

apps/mobile/src/screens/GroupWelcomePage/GroupWelcomePage.js
```diff
@@ -0,0 +1,28 @@
...
+      className='flex-1 bg-background'
+      contentContainerStyle={{
+        paddingTop: insets.top,
+        paddingBottom: insets.bottom,
+        paddingLeft: insets.left,
+        paddingRight: insets.right
+      }}
+    >
+      <View className='p-4'>
+        <HyloHTML
```
lorenjohnson, line 22: Did you get this going with <a> tags? It should work fine but you may need to look at other uses of HyloHTML and otherwise inspect the component itself to make sure you're giving what it needs to handle the links?

thomasgwatson: sooo, doesn't look like any instances of HyloHTML do any different styling beyond one instance of passing in a marginBottom. The base component does seem to have a set of default styles passed in: `import { baseStyle, tagsStyles, classesStyles } from 'components/HyloHTML/HyloHTML.styles'` So maybe something is missing from there? I'm on the road between LA and SF right now, so just pulled over to get this branch up to date with yours

lorenjohnson: I am not aware of what the issue is, you had just mentioned that <a>nchor tags were not working but wasn't sure what else you were seeing or experiencing. It sounds like something isn't doing what you expected here, but yes the styling is fairly fixed and not easy to override on a case-by-case basis. HTML in Mobile is a weird thing. So not sure you're flow, but either explore further and/or give me notes as to what is missing and/or quick instructions on how to test. Will add to todo backlog, whatever is there after release goes into tickets.

PR #410, opened 2025-03-15
feat(mobile): Applies derived Web background color, fixed Stream header

lorenjohnson:
This derives the 2-layered background color on Web and applies it as the default background across the app, feels much more like Web and I think is a worthwhile iteration towards alignment with the Web design. The only issue is the background color on the HyloEditorMobile I can't seem to change so the PostEditor in the app doesn't look quite right with this update. I'm overextended for tonight so gonna stop here, and it was strangely difficult to try and get that background color to go away in the editor, but certainly it's solvable and otherwise this update I think would be a notable improvement on the release... 

- Also makes the Stream header fixed at top of stream and gives it a bottom shadow to make it look right. Definite improvement.
- Install TestFlight build 5116 to take a look.


tibetsprague (approved):
amazing!

lorenjohnson:
@thomasgwatson why was this merged?

PR #418, opened 2025-03-15
Mobile post editor update

lorenjohnson:


lorenjohnson:
FYI -- Most of the delete here is in within the mobile PostEditor component and that file you need to manually trigger to look at the diff as its a large diff.

apps/mobile/src/screens/PostEditor/PostEditor.styles.js
```diff
@@ -1,4 +1,5 @@
 import { StyleSheet } from 'react-native'
+import { twBackground } from 'style/colors'
```
thomasgwatson, line 2: should have mentioned this yesterday but slight preference for something like `themeBackground` here, since its not about tailwind perse

thomasgwatson (approved):
lgtm

apps/mobile/src/screens/PostEditor/PostEditor.styles.js
lorenjohnson: I probably agree with you there, but either way this is a one-time'r to get us through this release and not meant to be establishing (and very much would not want it to establish) any sort of new pattern if that makes sense. You did see the note in colors.js about how I got to this background color form the tailwind theme yes? Due to transparency values it actually had to be computed manually...

PR #420, opened 2025-03-15
Ensure currentUser is passed into getStaticContext

thomasgwatson:
Was stopping the profile url from being correct


lorenjohnson (approved):
Great! This was on my list and quite happy to take it off now. Does this also get is back the avatar for current user in the header when on my?

lorenjohnson:
Answered my own question: Yes! Great.

PR #423, opened 2025-03-15
Mobile: Fixes default group banner not working, Adds /groups/:groupSlug/about

lorenjohnson:


thomasgwatson (approved):
So this is just to point /about at something, while we sort it out?

PR #424, opened 2025-03-16
chore(web): Makes center column scrollable again in WebViews (settings pages)

lorenjohnson:
In Mobile WebView usage the Group Settings areas are not scrollable without overflow being set to scroll for the center column. This change to hide the overflow was added in [48dab6b](https://github.com/Hylozoic/hylo/commit/48dad6be5b2aef1762d9cfd17c1a9ebd535afe0f) which was part of a changeset to fix the recently appearing new white extra bar on the right side of all WebView pages. Unfortunately that changeset did not change the appearance in the app in these WebViews, the white bar is still there and went unchanged. Reverting this one setting for now makes things not broken, so think we should move this forward and address the larger issue as another pass.

See the white bar in the screen shots. They appear with or without this overflow setting set to scroll:

Simulator Screenshot - iPhone 16 Pro Max - 2025-03-16 at 01 57 41

Simulator Screenshot - iPhone 16 Pro Max - 2025-03-16 at 01 56 21

### DIFF
--- a/apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.module.scss
+++ b/apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.module.scss
@@ -328,7 +328,7 @@ $drawer-z-index: 51;
   min-height: 1px; // Prevent collapsing
   height: 100%;
   flex: 1;
-  overflow: scroll;
+  overflow: auto;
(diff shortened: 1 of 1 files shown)


tibetsprague (commented):
its slightly more complex than this, i just reverted my change that broke this

PR #426, opened 2025-03-16
chore(mobile): Fixes Android deep linking configuration

lorenjohnson:
Not entirely sure it was broken before, but these changes are all clarifying and fixes the open-link helper as well as confirms deep linking is now working in Android.


thomasgwatson (approved):
lgtm

PR #445, opened 2025-03-19
Dropzone TLC

### LINKED ISSUE #415: Dropzones when editing menu are all kinds of weird
- [x] We shouldnt be able to put containers inside other containers
- [ ] Sometimes the drop zone i want is missing
- [ ] Other times the wrong ones are there
- [ ] ~~Often they dont appear in a way that corresponds to where my mouse is, especially when dragging a big container. This is confusing.~~ [This is a known issue for dnd-kit](https://github.com/clauderic/dnd-kit/issues/1432) and a new release is planned. The new release will require migration :b
- [x] Hitting ESC while dragging shouldn't make the widget disappear (use case: hit ESC to cancel drag and put widget back)

What Tom sees:

- [x] non-container widgets are no longer including a dropzone, so you can't place things directly above them (BUT ONLY WHEN THE ACTIVE WIDGET IS NOT A CHILD WIDGET)
- [ ] ~~escape key also needs to preventDefault I think, seemed to impact my browser hotkeys for escape (bailed from full-screen)~~ The library calls event.preventDefault() but doesn't call event.stopPropagation(). Not only that, but the event we receive is a `PointerEvent` so it's not the keyboard event :b
- [x] Protected containers are not protected because the bottom-of-children dropzone is still being inserted; _could move a non-chat widget into chats_
- [x] ~~Removal dropzone not visible unless something is dragged over it; impossible to find~~ Tibet said he wants to put a 'delete' button on widgets in edit mode.

KevinTriplett:
closes #415 

Work so far:

- Pressing ESC cancels the drag and puts the widget back


KevinTriplett:
I'm thinking to experiment with the "Sortable" feature: https://docs.dndkit.com/presets/sortable

It depends on whether or not we can still drop widgets outside of containers like we currently do. I think we can and it's more what a user is expecting inside containers, imo.

tibetsprague:
nice. I do plan to add a delete button to each widget when in edit mode which will also be more obvious than dragging it off

KevinTriplett:
> nice. I do plan to add a delete button to each widget when in edit mode which will also be more obvious than dragging it off

Kewl! Yeah, that would be a great improvement

KevinTriplett:
I think this "disallow container drop on container" may have fixed your wild dropzones, although there's one missing dropzone at the bottom of the menu, I'm going to add it now.

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -461,10 +470,21 @@ function GrabMe ({ children, ...props }) {
   )
 }
 
-function DropZone ({ droppableParams, isDroppable = true, height = '', hide = false, children, removalDropZone }) {
+function DropZone ({ droppableParams, isDroppable = true, hide = false, children, removalDropZone }) {
   const { setNodeRef, isOver } = useDroppable(droppableParams)
+  const { contextWidgets, activeWidget } = useContextMenuContext()
+  const activeWidgetIsContainer = ['container', 'home', 'chats', 'members', 'setup'].includes(activeWidget?.type)
+    || activeWidget?.childWidgets?.length > 0
+  const isIllegalDrop = droppableParams.id.includes('bottom-of-child-list') && activeWidgetIsContainer
```
thomasgwatson, line 478: This is likely duplicating logic in `isValidChildWidget`.

Given the product and implementation context of the drag and drop stuff, I think it might be best for me to dig into this issue. I would have assigned myself earlier in the week but we've been deep in rare in-person time.

thomasgwatson:
@KevinTriplett  I've added a few more specific issues I've spotted to the original ticket. Do you want to pair on this at some point in the next day or two?

KevinTriplett:
> @KevinTriplett I've added a few more specific issues I've spotted to the original ticket. Do you want to pair on this at some point in the next day or two?

Sure! Here's my calendly link, it'll help us figure out time zones: https://calendly.com/kevintriplett/90-minute-project-meeting

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
KevinTriplett: Yeah, it's a bit complex, this seems to work but let's take a look at it and see if we can avoid dups

KevinTriplett:
I updated the issue as well -- I see the isValidChild function. Let's see if we can combine the two, there are things I added that are not in the function. It would be good if they were all in the function. There are some things we can't do because the library doesn't give us the keyboard events and Tibet wants to remove widgets in a different way, which I think would be great.

KevinTriplett:
> Ahh sad about the escape keystroke inaccessibility :( I booked a time for this afternoon for us to do a session, let me know if that works for you :)

I know, right? Looking forward to working with you in a couple hours.

KevinTriplett:
Oh, and I pushed a commit that you can pull and try out, have a look at. There are some questions and one puzzle, in addition to hitting the rest of the issue list.

KevinTriplett:
@tibetsprague pls review -- one last issue that @thomasgwatson will take: although there's a dropzone at the bottom of a child list, the backend is not adding it to the end of the parent. I'll create a separate issue for Tom to take.

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -497,37 +516,38 @@ function DropZone ({ droppableParams, isDroppable = true, height = '', hide = fa
...
 
-  if (activeWidget && activeWidget.id === item.id) {
-    hideDropZone = true
+  // don't render the active widget, unless it's the overlay
+  if (!isOverlay && activeWidget && activeWidget.id === item.id) {
     return null
   }
 
-  const isActive = item.viewUser?.lastActiveAt ? new Date(parseInt(item.viewUser.lastActiveAt)) > new Date(Date.now() - 1000 * 60 * 4) : false
+  // XXX why is this using parseInt? convert to luxon?
```
thomasgwatson, line 545: Were you satisfied with my explanation Kevin?

thomasgwatson (approved):
lgtm

packages/presenters/ContextWidgetPresenter.js
```diff
@@ -150,23 +150,59 @@ export function getStaticMenuWidgets ({ isPublicContext, isMyContext, profileUrl
...
 
+export function isValidDropZone ({ overWidget, activeWidget, parentWidget, isOverlay = false, isEditing, droppableParams }) {
+  const containerTypes = ['container', 'custom-views', 'chats', 'members', 'setup', 'auto-view', 'home']
+  const isWrongType = overWidget?.type === 'home'
+    || (parentWidget?.type === 'chats' && !activeWidget?.viewChat?.id)
+    || (parentWidget?.type === 'custom-views' && !activeWidget?.customView?.id)
+  const parentWidgetIsContainer = containerTypes.includes(parentWidget?.type)
+  const activeWidgetIsContainer = containerTypes.includes(activeWidget?.type)
+  
+  // const listBottom = droppableParams.id?.includes('bottom-of-child-list')
```
tibetsprague, line 184: remove?

tibetsprague (approved):
havent tested yet but code seems fine

Issue #449, opened 2025-03-19
Group explorer is a 404 in the app


tibetsprague:

lorenjohnson:
dupe of #440

thomasgwatson:
I regret to inform you that BOTH the Group Explore Page and the Group ExploreR Page need to be surfaced on the app :(

lorenjohnson:
Got it. I was nearby so just corrected that... :)

Issue #440, opened 2025-03-18
Group Explore WebView now seems to forward out of Group Explore to another 404 page.


thomasgwatson:
Group Explore WebView now seems to forward out of Group Explore to another 404 page. Group Explore is currently plumbed in Mobile to be a destination to view group that you are not a member of, but I suspect this redirect is because that isn't actually proper use. So the solution here may be to just decide what the proper destination is for a modal pop-up to preview a group you're not a member of...

PR #450, opened 2025-03-19
chore(mobile): Link to Public > Group Explorer, WebView clean-up

lorenjohnson:
- Adds Public > Group Explorer page via WebView #449 
- Adds updated URL / handling of Group Explore modal for Group Explorer WebView and Groups within Groups #440 
- Normalizes some/all WebView routing capture back into App. In the past this was done on a case-by-case basis in anticipation of centralizing some or all of it. After looking at it a bit it seems we're now in a place to do that, so doing that here. It is worth some testing, but I went through it moderately through both in thinking through the code logic and in manual testing, and find no issue.
- There is likely one or two more steps of simplification which can be done here that I'll save for another pass in the future (e.g. `UserSettingsWebView` can probably be disposed of in lieu of direct `HyloWebView` usage, and `nativeRouteHandler` maybe could/should also be centralized...)


thomasgwatson (approved):
lgtm

PR #468, opened 2025-03-21
create DateTimeHelper util file

### LINKED ISSUE #466: Create a shared luxon helper file
Purpose: simplify the code by creating a DateTimeHelper.js file, especially around creation and formatting.

KevinTriplett:
closes #466 

Tibet, please test the TextHelper functions, looks like you recently changed these to get emails correct.

Main purpose is to wrap luxon `DateTime.from*` with general purpose conversion of string and JS Date objects, plus add locale and timezone support in one easy function, to reduce code noise and errors of omission.

Cons are that we're not using the library directly. Pros are that it's nice not to have to think about whether the input date is a string or JS Date object.


lorenjohnson:
Hi @KevinTriplett! Jumping in here with a couple of code-level organization notes:

- For consistency, please pluralize this helper object as `DateTimeHelpers` and export the module like the other helpers by adding the import/export in `packages/shared/index.js`. Then in use `import { DateTimeHelpers } from '@hylo/shared'`. 
- There are some other date/time helpers already in place in `TextHelpers`, please move these to `DateTimeHelpers` and update references accordingly. Some of those will be in Mobile and Backend so make sure those are confirmed working (Have you been setup to run Mobile?)
- Ideally it would be good to reconcile / use these new helpers in Mobile and Backend as well, replacing current methods, where or if appropriate.
- _FYI - I have a personal preference for these helpers to be always used with by only importing the helper module and not destructing the composite methods at the time of import. So that means using methods like `DateTimeHelpers.from*()` vs `from*()`. I find that additional clarity and context helpful for my head, and is something I do consistently in the Mobile code base, however it is not by any means something we as a team have agreed to one way or another, so either way is fine._

KevinTriplett:
Hi Loren, I look forward to meeting you someday. I need to get setup for mobile, I'd like that, can you point me to next steps to get setup?

I see where you're going with the whole import the module thing, it does add clarity. I would be happy to do that for both helper files if the team agrees to it. Can we (or you) all have that conversation? And when does the dev team regularly meet up now?

THanks for checking on this!

KevinTriplett:
Ah! I see that's already been done with TextHelpers, importing without destructuring. Okay, I think that's good enough to go with doing it for DateTimeHelpers, it's a precedent. COngrats, I'm helping you move this toward standard practice.

thomasgwatson:
Since I'm in California for a few weeks, we have a bit of time with much simpler dev team scheduling. @tibetsprague whats a good time? We could do earlier monday Pacific time, so its earlier in the Berlin evening?

lorenjohnson:
HI again @KevinTriplett! Yes, I look forward to meeting sometime too. Maybe a video call soon to meet/greet :) Your work is really appreciated!  Your additional updates here looks good to me. There are some _maybe somewhat_ update-to-date instructions on getting going in Mobile dev in the README but yes probably having one of Tom, Tibet, or I support you through that may be best. 

I think in this case it would be good to tag the upcoming Mobile release Milestone in this PR (6.0.1) such that when that release is being readied and tested we know to look to the spots where there were changes here. I see there are some merge conflicts to handle as well.

lorenjohnson (approved):
Not wanting to block any work moving forward here:) --- I haven't tested, but generally in reviewing the changes since my comment things look good, and I imagine this is fine to move forward once the merge conflicts are resolved.

PR #476, opened 2025-03-21
Sync mobile stream with user settings for stream

### LINKED ISSUE #444: Look into saving Stream settings on user, it is currently only incompletely implemented on Mobile
NOTE: there are some stream settings that don't apply to mobile, because they reference different viewModes that don't exist on mobile

Look into saving Stream settings on user, it is currently only incompletely implemented on Mobile

thomasgwatson:
Closes #444


lorenjohnson (approved):
- Please update that small bit according to the one review comment. 
- The values for filter and sortBy in this setup won't be updated if the selected customView or currentUser changes while this screen is still mounted. I can't see a case where this ever is an issue with currentUser, but does this never happen currently with customView? If not, I guess we're good, but if so then we need a useEffect to keep these initial values from being stale.

PR #486, opened 2025-03-24
408 preview auto plays

### LINKED ISSUE #408: Going to chat room where there is a post card with a featured youtube video immediately opens up the youtube app
There's no way for me to stop that and stay in Hylo

thomasgwatson:
Basically this stops the use of react-player for webViews in mobile. Instead it renders a regular preview. Closes #408

I tested this on Android and encountered an unrelated issue, where we hadn't escaped the use of KeyboardManager (iOS only) for Android users.

Screenshot 2025-03-24 at 11 04 28 AM

### DIFF
--- a/apps/mobile/src/screens/GroupWelcome/GroupWelcome.js
+++ b/apps/mobile/src/screens/GroupWelcome/GroupWelcome.js
@@ -21,6 +21,7 @@ import GroupWelcomeTabBar from 'screens/GroupWelcome/GroupWelcomeTabBar'
 import HyloHTML from 'components/HyloHTML'
 import Pill from 'components/Pill'
 import styles from 'screens/GroupWelcome/GroupWelcome.styles'
+import { isIOS } from 'util/platform'
 import { caribbeanGreen } from 'style/colors'
 
 export const addSkillMutation = gql`
@@ -72,9 +73,9 @@ export default function GroupWelcome () {
   const [allQuestionsAnswered, setAllQuestionsAnswered] = useState(!currentGroup?.settings?.askJoinQuestions || !!joinQuestionsAnsweredAt)
 
   useEffect(() => {
-    KeyboardManager.setEnable(true)
+    if (isIOS) KeyboardManager.setEnable(true)
     return () => {
-      KeyboardManager.setEnable(false)
+      if (isIOS) KeyboardManager.setEnable(false)
     }
   }, [])
 
--- a/apps/web/src/routes/ChatRoom/ChatPost/ChatPost.js
+++ b/apps/web/src/routes/ChatRoom/ChatPost/ChatPost.js
@@ -304,8 +304,11 @@ export default function ChatPost ({
           delay={250}
           id='flag-tt'
         />
-        {linkPreview?.url && linkPreviewFeatured && isVideo && (
+        {linkPreview?.url && linkPreviewFeatured && isVideo && !isWebView() && (
           <Feature url={linkPreview.url} />
+        )}
+        {isWebView() && linkPreviewFeatured && (
+          <LinkPreview {...pick(['title', 'description', 'imageUrl', 'url'], linkPreview)} className={styles.linkPreview} />
         )}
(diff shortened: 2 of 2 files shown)


apps/mobile/src/screens/GroupWelcome/GroupWelcome.js
```diff
@@ -72,9 +73,9 @@ export default function GroupWelcome () {
   const [allQuestionsAnswered, setAllQuestionsAnswered] = useState(!currentGroup?.settings?.askJoinQuestions || !!joinQuestionsAnsweredAt)
 
   useEffect(() => {
-    KeyboardManager.setEnable(true)
+    if (isIOS) KeyboardManager.setEnable(true)
```
tibetsprague, line 76: what is this about?

tibetsprague (approved):
good fix for now, but would be real nice to be able to play YouTube videos in the stream eventually! there must be a way

apps/mobile/src/screens/GroupWelcome/GroupWelcome.js
thomasgwatson: > I tested this on Android and encountered an unrelated issue, where we hadn't escaped the use of KeyboardManager (iOS only) for Android users.

lorenjohnson:
Good catch on the KeyboardManager bit, indeed that component is iOS only.

As for React Player, I think what you did is a tidy and reliable solution, so am ok with it and it may be the most reliable solution for the moment. 

However, I did look into it a bit further as it would be nice if the Chat continued to work with inline video playing, and generally I was quite curious about the underlying issue was that was keeping videos from playing as expected in Chat as there is no WebView or mobile browser limitation on HTML5 players or iFrames that I know of.  I was able to get the video player to work as expected by adding to the originWhiltelist in our WebView usage. Here is that patch:

https://github.com/Hylozoic/hylo/tree/mobile-408-fix-react-player-in-webview

**Some background and caveats:** 

In WebViews we are always keeping the user from ending-up on a page other than the one we targeted. Basically if we open the Chat view for a group, we want to make sure the user stays on the Chat page and that we take control of what happens when there is an attempt to navigate to another page within the Web app or externally. That is probably already clear enough, but I have been wanting a place to explain this a bit more as it is the only consequential complexity / trickiness specific to WebViews that isn't just a general mobile/responsive issue is found in here and in the various means this is addressed. 

This is a weird case, as it is not at all expected to me that in loading an HTML5 player and/or iFrame that the URLs used in those components would trigger the WebView "the user is linking to something" route handling. After looking into it a bit deeper and digging down the layers though, it also _kinda_ maybe makes sense. With a whiteboard and 20 minutes I could explain why :) 

Regardless, the solution here to make the WebView not do this weird thing with the HTML5 Player and iFrame instances generated by ReactPlayer is to whitelist the host origins of the sources that we expect to appear there. In that branch I just update the originWhitelist for the HyloWebView and it seems to make everything work as expected. Take a look and test. As long as this also works in Android, which is should, then this will keep existing functionality and avoid another isWebView() exception in Web code, which is something we always worth avoiding unless absolutely necessary.

Note that in my solution I am whitelisting only YouTube, Vimeo, and Soundcloud and have only tested in iOS for each of those 3 embeds. I have not tested for a Facebook video, or for a handful of other lesser known services (Wistia, etc) which may bring up this issue if not also whitelisted. 

_If we go with the isWebView() limited functionality route of this PR, then I'd recommend some note is added as a code comment above where the isWebView() call had been added, or in the HyloWebView component, or both._

thomasgwatson:
Oh great! I tried messing with some of the other props in the webview but not this one.

lorenjohnson:
@thomasgwatson please do though make sure to test this yourself and in particular give it a whirl in Android, as I didn't test there. If it's to your satisfaction, great, but if not it could still be that your more conservative take has merit for now...

thomasgwatson:
Oh yeah, I got it running on Android and I don't think there are more sites
we need to add to the whitelist

PR #505, opened 2025-03-26
Update queryParamWhiteList

### LINKED ISSUE #480: #Topic Stream not filtering correctly
Device: Macbook / Chrome

The Topic Streams aren't filtering correctly. Navigate to the All Topics view (in Building Hylo) and click on a topic, all the posts shared show up. I clicked on various topics and experienced the same issue.

I would expect to see posts/content filtered by that topic.

https://hylo.canny.io/admin/board/bugs-feature-requests/p/topic-stream-not-filtering-correctly

thomasgwatson:
Closes #480

### DIFF
--- a/apps/web/src/store/reducers/queryResults.js
+++ b/apps/web/src/store/reducers/queryResults.js
@@ -337,6 +337,7 @@ export const queryParamWhitelist = [
   'search',
   'slug',
   'sortBy',
+  'topic',
   'topics',
   'type', // TODO: why do we have type & filter? should only need one
   'types',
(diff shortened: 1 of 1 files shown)


tibetsprague (commented):
weird. so chat room uses topic so I think this would break chat rooms. I guess we can have both? or switch the topic stream to just use topic like chat room.

thomasgwatson:
Ahh right. It still worked for the initial load (which is does for similar situations when a param is missing from the list but included in the actual request) but it failed when `hasMore`-ing

Added topic back in.

This will all go away (and we'll have a different set of annoying things to deal with) once we urql this shizz

tibetsprague (approved):

Issue #439, opened 2025-03-18
Topic links navigate to stream with empty content in My Stream views


thomasgwatson:
Topic links navigate to stream with empty content in My Stream views

PR #507, opened 2025-03-27
All topics route is respected; topics navigate to stream; stream resp…

### LINKED ISSUE #506: Add all topics screen and/or routing to mobile
All topics widget was added to web and backend but doesn't route to anything on mobile

thomasgwatson:
…ects topicName; stream respects all-post filter

Closes #506 #439


apps/mobile/src/navigation/HomeNavigator.js
```diff
@@ -57,6 +56,7 @@ export default function HomeNavigator () {
   return (
     <HomeTab.Navigator {...navigatorProps}>
       {/* WebView screens (may link/route internally) */}
+      <HomeTab.Screen name='All Topics' component={HyloWebView} />
```
thomasgwatson, line 59: Thought I'd give HyloWebView a spin here

apps/mobile/src/screens/Stream/Stream.js
```diff
@@ -186,7 +189,7 @@ export default function Stream () {
 
   const handleFilterChange = (filter) => {
     setFilter(filter)
-    updateUserSettings({ changes: { settings: { streamPostType: filter } } })
+    updateUserSettings({ changes: { settings: { streamPostType: filter || null } } })
```
thomasgwatson, line 192: Mutation didn't like `undefined` as an input

apps/web/src/routes/FullPageModal/FullPageModal.js
```diff
@@ -20,8 +20,8 @@ export default function FullPageModal ({
               key={tab.path}
             />
           ))}
-          {!multipleTabs && (content || children)}
         </Routes>
+        {!multipleTabs && (content || children)}
```
thomasgwatson, line 24: In All Topics case, there are not multiple tabs, and thus, it was trying to render `children`, which was a div. And <Routes> only accepts <Route> as a valid child and it was flipping out

apps/mobile/src/components/HyloWebView/HyloWebView.js
```diff
@@ -40,6 +40,9 @@ export const useNativeRouteHandler = () => {
       navigation.navigate('Stream', { topicName })
     },
     '(.*)/topics/:topicName': ({ routeParams: { topicName } }) => {
+      navigation.navigate('Stream', { topicName })
+    },
+    '(.*)/chats/:topicName': ({ routeParams: { topicName } }) => {
```
tibetsprague, line 45: wait, how was this working before?

apps/mobile/src/navigation/linking/index.js
```diff
@@ -128,8 +128,9 @@ export const routingConfig = {
   '/:context(groups)/:groupSlug/members':                                 `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Members`,
   '/:context(groups)/:groupSlug/members/:id':                             `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Member`,
   // TODO: Routing -- Actually legacy redirection, consider creating or adding to a redirect mapper
-  '/:context(groups)/:groupSlug/topics/:topicName':                       `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Chat Room`,
```
tibetsprague, line 131: can you explain this change to me?

tibetsprague (approved):
some questions, more just for me to understand

apps/mobile/src/components/HyloWebView/HyloWebView.js
lorenjohnson: This handler is used to change the behavior of what happens when a link is pressed in the WebView if we want it to be other than the default openURL behavior.  The special cases here were all recently consolidated from a few different WebViews into this central place, but I think with simplifications I've made to linking, routing, openURL that most if not all special cases can be removed and left to the default case. Again, the exceptions should just be when we want the link pressed in the WebView to do something different than if the user had pressed on the link in the app or outside the app to be opened in the app... Check me on that, as don't have the time to jump in and test it myself at the moment, but I think you'll find that is true. If it is this would be a good time to get clear on that and take out any unnecessary special cases.... Hope that makes sense.

apps/mobile/src/navigation/HomeNavigator.js
lorenjohnson: For now  I've done away with having special components or Screen entries for WebViews that don't have any special handling, and instead just direct the route in linking/index to `${AUTH_ROOT_SCREEN_NAME}/Drawer/Tabs/Home Tab/Web View`. I'd do the same for this as it currently has no special handling.

See https://github.com/Hylozoic/hylo/blob/3cd1b22ab6e0f861237ea56713cd43a1b13c8ab2/apps/mobile/src/navigation/linking/index.js#L63 and https://github.com/Hylozoic/hylo/blob/3cd1b22ab6e0f861237ea56713cd43a1b13c8ab2/apps/mobile/src/navigation/linking/index.js#L116.

lorenjohnson (changes requested):
Good work digging into WebView usage a bit more 👍 A couple small things to consider/change, but otherwise looks good to me.

apps/mobile/src/navigation/linking/index.js
lorenjohnson: I don't know what is correct, and this is probably apparent here... but this means that `/:context(groups)/:groupSlug/topics/:topicName' routing now goes to the Stream instead of Chat, which I assume is correct now. If so please wondering if the TODO still is correct or if not if it should be removed?

apps/mobile/src/components/HyloWebView/HyloWebView.js
thomasgwatson: The main functional outcome is that /topics/:topicName was pointing at the ChatRoom, not the Stream.

apps/mobile/src/navigation/linking/index.js
thomasgwatson: Yep, /chat goes to chat and /topics goes to Stream

PR #517, opened 2025-03-28
Reduce stream page size; Reduce number of commenter Avatar urls reque…

### LINKED ISSUE #513: Stream loading very slowly and usually not at all for me in mobile app
The home chat loads fine but the stream takes more than 10 seconds to load for most groups, and sometimes just seems to spin forever

thomasgwatson:
…sted; remove redundant query elements from postQuery; comment out badge code from postHeader

Closes #513


thomasgwatson:
Its so hard to tell on local dev, since there is no real server round-trip. Once we get a build up on test flight I'll get you to try it out

apps/mobile/src/components/PostCard/PostHeader/PostHeader.js
```diff
@@ -41,14 +41,14 @@ export default function PostHeader ({
   const [{ currentGroup }] = useCurrentGroup()
   const handleShowMember = () => showMember && showMember(creator.id)
 
-  const creatorHasResponsibility = useHasResponsibility({ person: creator, forCurrentGroup: true })
+  // const creatorHasResponsibility = useHasResponsibility({ person: creator, forCurrentGroup: true })
```
lorenjohnson, line 44: Please remove these and all remaining commented-out code if it is not longer used.

packages/graphql/fragments/postFieldsFragment.js
```diff
@@ -134,12 +110,6 @@ export default gql`
       position
       id
     }
-    postMemberships {
```
lorenjohnson, line 137: I would keep this for now as this fragment is used throughout, and it has caching benefits to query a post using the same fragment as much as possible... If it is making a non-trivial difference in the query time to remove this postMemberships id / group.id part then it is worth looking at more carefully, but my suspicion that this particular part isn't changing these substantively. You can try URQL dev tools to get a read on the actual query times in dev to compare.

lorenjohnson (changes requested):

packages/graphql/fragments/postFieldsFragment.js
tibetsprague: but postMemberships are not used anywhere in the code base

packages/graphql/fragments/postFieldsFragment.js
```diff
@@ -110,6 +110,12 @@ export default gql`
       position
       id
     }
+    postMemberships {
```
tibetsprague, line 113: noo, kill with fire. we dont use them

packages/graphql/fragments/postFieldsFragment.js
thomasgwatson: Loren, does this get used by urql to key a relationship between posts and memberships that we want to track? Doing a quick scan of the monorepo, it doesn't seem to be used explicitly

lorenjohnson: I guess not, totally good removing them then.

lorenjohnson (approved):

PR #518, opened 2025-03-29
516 layout pain in webview

### LINKED ISSUE #516: Extra scroll bar on the right side in mobile app chat rooms
Can't reproduce on web at all. 
It becomes extra problematic when the cursor is in the chat box, then if you try to scroll the page it scrolls to a huge empty space under the chat box

thomasgwatson:
Closes #516 

Had to tweak the padding on the chat contents, as it was off-center and uneven (adjusting for the presence of a scroll-bar perhaps)

Screenshot 2025-03-28 at 6 58 18 PM


apps/mobile/src/components/HyloWebView/HyloWebView.js
```diff
@@ -77,6 +77,32 @@ const handledWebRoutesJavascriptCreator = loadedPath => allowRoutesParam => {
...
+  
+  body {
+    width: 100vw !important;
+    position: fixed !important;
+    left: 0 !important;
+    right: 0 !important;
+    max-width: 100% !important;
+  }
+`
+
```
lorenjohnson, line 105: Good you've got to the bottom of this! Tell me more on this part? I'm dubious that this part is WebView specific. WebKit specific for sure, which the first rule of course is explicitly. If I remember right I used to have a very similar set of rules set in Web as the issue of the scrollbar taking up space even when no overflow in Safari is a familiar issue to me. Either way, I'm not convinced that there is something WebView specific going on here. If so I'd like to understand what it is and why and to make a note of it here in a comment, and if not I'd then these should be applied to Web to address what I assume are WebKit / Safari issues on mobile devices and perhaps also on Desktop...

apps/mobile/src/components/HyloWebView/HyloWebView.js
```diff
@@ -149,6 +175,7 @@ const HyloWebView = React.forwardRef(({
         window.HyloWebView = true;
         ${path && handledWebRoutesJavascriptCreator(path)(handledWebRoutes)}
       `}
+      customStyle={`${injectedCSS}`}
```
lorenjohnson, line 178: `customStyle` when set in HyloWebView usage, as it is [here](https://github.com/Hylozoic/hylo/blob/0322e320074acd3e2301c8209052932996c1a174/apps/mobile/src/components/HyloEditorWebView/HyloEditorWebView.js#L133), will overwrite this. If this CSS needs to be HyloWebView specific CSS for all Hylo Web pages then I'd either destructure `customStyle` from incoming props and then concat it this `injectedCSS` (which I'd call `defaultCustomStyle` to not add any unnecessary namespace complexity), or if it is a Chat specific need I'd apply this as the `customStyle` prop for the Chat WebView use of HyloWebView.

lorenjohnson (changes requested):

apps/mobile/src/components/HyloWebView/HyloWebView.js
thomasgwatson: Yeah, agreed that its not webview specific and could be applied elsewhere. I did try a version of applying this in web AuthRouter but that either didn't have the right impact (although I was using a much narrower set of style changes) or its in the wrong part of the component/styling hierarchy.

Somewhat terrified of adding a whole set of rules to the whole web app and then unintentionally breaking scrolling in a bunch of places. So felt safer to try at the 'inject into webview' level and assume Web styling is handling safari/browser differences on its own already.

apps/mobile/src/components/HyloWebView/HyloWebView.js
thomasgwatson: The issue was showing up on other webviews as well (group explorer for one). I think its not obvious on settings pages, just due to their layout/styling. Could certainly make this propable

lorenjohnson: I gotcha this and above ^^. So if we want to keep it here for now and not figure-out how to adjust the Web stylesheet for this, then please rename `injectedCSS` to `baseCustomStyle` (better than `default` actually as these will always be applied, not replaced by the incoming `customStyle` prop). Then de-structure `customStyle` out of the incoming props as `providedCustomStyle`and concat that into an in-component const of `customStyle` which then gets set to that prop on the WebView. In terms of naming that is how I'd handle this and is inline with similar cases here in the Mobile code. Either way though the result we want are these base style overrides always applied in `HyloWebView` while still allowing other overrides from the uses of `HyloWebView` upstream to also be applied after them. Do you follow, and does that make sense to you too?

lorenjohnson: Also, maybe add a `// Should probably just be applied to Hylo Web stylesheet as what this solves is not necessarily WebView specific, but putting here for now to limit possible untested impact on Hylo Web generally" comment or similar above the `const baseCustomStyle` defintion.

lorenjohnson (approved):
Please make suggested amendments or let me know first we're seeing something different and why, and you have my approval with either action.

tibetsprague:
this breaks other layout stuff 

Screenshot 2025-03-30 at 1 47 14 PM

PR #521, opened 2025-03-31
Revert "PR #468 from Hylozoic/466-create-datetime-helper-file"

lorenjohnson:
Reverts https://github.com/Hylozoic/hylo/pull/468 which was not prepared for Mobile.

PR #522, opened 2025-03-31
Move to 6.0.2-0 prerelease

lorenjohnson:
Updates Mobile version to next pre-release version (6.0.2-0) as per Mobile README release instructions. Once a version is released Apple will reject any further submissions under that version number, so this allows builds to happen again. 

Base branch is https://github.com/Hylozoic/hylo/pull/521 which reverts issues introduced by #468 `@hylo/shared/DateTimeHelper` work which is was blocking successful builds of Mobile.

BitRise builds https://app.bitrise.io/build/6dec3223-cafe-4030-aa43-dc40e2acaa04 (Android) and https://app.bitrise.io/build/14b5a0e3-d4d0-4ac0-85a5-d86017452030 (iOS) build this branch against Production API to allow testing of Stream query changes, etc.

### DIFF
--- a/apps/mobile/README.md
+++ b/apps/mobile/README.md
@@ -66,7 +66,7 @@ Here are the common scenarios in which we should run the `bump-version` command:
   - Upload the APK file downloaded from Bitrise and tested above
   - Add release notes
   - Submit for review/release
-- Once the release is accepted by both the stores run `yarn bump-version version prereleaase` on `dev`, commit the changes, then `git push --tags` to setup the next prerelease build versioning.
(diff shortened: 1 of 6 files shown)


lorenjohnson:
`yarn bump-version prepatch` ... but you'll currently find that the `apps/mobile/ios/project.pbxproj` has some lines in it which offend the underlying react-native-version libraries parser so it that command will fail without temporarily removing those lines... Will file an issue about that, but you would have encountered it if you had tried recently.  Just a reminder that a manual update requires updating both the iOS and Android project files in addition to the `apps/mobile/package.json` of course (as you will see was done by the `bump-version` script in this branch).

PR #523, opened 2025-03-31
519 post layout responsive reset

### LINKED ISSUE #519: Layout of posts in chat room broken
Image

thomasgwatson:
Closes #519 

Chat examples at two breakpoints and a look at Stream to ensure its still happy as well

Screenshot 2025-03-31 at 9 13 32 AM
Screenshot 2025-03-31 at 9 23 33 AM
Screenshot 2025-03-31 at 9 23 43 AM


tibetsprague:
Great I just want to test on my machine

tibetsprague:
so chats are still not lined up correctly on the left they should look like production with the content lined up with the name
Screenshot 2025-03-31 at 8 10 37 PM

tibetsprague:
theres also now an extra bit of padding on the right, even on chrome
Screenshot 2025-03-31 at 8 12 14 PM

thomasgwatson:
Yeah, that is intentional.

Makes the padding even on both sides

Screenshot 2025-03-31 at 8 18 44 PM
Screenshot 2025-03-31 at 8 18 58 PM

tibetsprague:
ok, im cool with that, though i could see tweaking it more to maximize space on mobile

thomasgwatson:
Mmmmm, I've just had a little dig, and there doesn't seem to be a straight-forward way to tighten up the spacing in the chat room. There are almost no margins or padding values to tweak.

Here you can see that the message is wrapped by some spaces, but that those are seemingly unassociated with the element, or its margin/padding
Screenshot 2025-04-01 at 7 03 06 PM

I tried to see if there was a parent container that had a flex gap or similar but didn't find one. The second element just has a computed top value, so its 156 pixels down from the section heading...

Had a scan of Virtuoso's docs and didn't find anything help about adjusting the space between items

tibetsprague:
@thomasgwatson can we please re-align the left of the text content of chat posts with the person's name 🙏🙏🙏

tibetsprague:
tightening of padding has to do with changing the padding of <p> tags within chat post content

Issue #504, opened 2025-03-26
After filtering map drawer the posts dont appear


tibetsprague:

thomasgwatson:
Drawer posts aren't showing up for me at all. The graphql action is returning posts on the group, so it seems like a selector issue... Have we made any chances to the drawer that might have tweaked a param or added something that isn't on the param whitelist? I've tried to scan parts of related code but haven't found any changes yet...

tibetsprague:
Yes, im working on this

tibetsprague:
it was the adding topics to the querySelector that you did that broke it

tibetsprague:
but there are a lot of issues with the map and drawer right now, im 80% through

tibetsprague:
sorry I should have put my name on this

thomasgwatson:
So, that means other post queries have to include `topics` as an empty param for the selector to work? Why was it dogging the drawer selector but the map post selector was happy? Well, I guess its all going to be paved over by urql in time anyway...

hylobot:
It’s tricky but easy fix, in the fetch we had set topics = null, and in the selector we had topics = []

PR #532, opened 2025-03-31
Fix display of posts in map drawer

tibetsprague:
Once again reload them after map filters change (like post types)

Also fix zoom level when jumping to a location on the map

Fixes https://github.com/Hylozoic/hylo/issues/504


thomasgwatson (approved):
lgtm

Issue #528, opened 2025-03-31
Weird loading animation in mobile app


tibetsprague:
Whenever i open the new Hylo app, after the spinner in the center of the screen it shows a blank screen with the name of the group at the top, then a drawer kind of slides over that is the exact same screen, blank with the name of the group at the top, before the page loads. Can we avoid that initial load + drawer slide over and just directly load the page we are trying to go to?

thomasgwatson:
dupe of #435

lorenjohnson:
Resolved with https://github.com/Hylozoic/hylo/pull/534

Issue #435, opened 2025-03-18
Double navigation bouncing


thomasgwatson:
Look into double screen navigation when linking into app, after join group, etc... Visible by the screen seeming to load twice and almost something to be fixed in the useHandleCurrentGroup area... Nothing breaks here, it's just clunky and mean the screen (most often a chat webview) is created then destroyed and created again which is not awesome

PR #534, opened 2025-04-01
493-mobile-group-context-fix

### LINKED ISSUE #493: Click on some Push notifications going to the wrong group
I have seen this sometimes, but not every time. When it does it takes me to some random group I have not visited in a long time

lorenjohnson:
- Fixes #493: ChatWebView was always using originalLinkingPath even when currentGroup.slug changed. Now it re-constructs path using currentGroup and the parsed path params
- Updates membership lastViewedAt on change to group, which seemed to be currently relying upon the update within WebView Chat
- Simplifies some HyloWebView implementation details for "handled routes" 
- Fixes currentGroup switch double-nav event (https://github.com/Hylozoic/hylo/issues/435 / https://github.com/Hylozoic/hylo/issues/528)
- Update GroupSettingsWebView to also not rely exclusively on originalLinkingPath to fix the same issue: currently when in group settings and then linking to another screen in another group, the settings screen will not update to the newly selected current group

### DIFF
--- a/apps/mobile/src/screens/GroupSettingsWebView/GroupSettingsWebView.js
+++ b/apps/mobile/src/screens/GroupSettingsWebView/GroupSettingsWebView.js
@@ -1,27 +1,18 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import React, { useCallback } from 'react'
 import { ScrollView, StyleSheet, Text, TouchableOpacity, View } from 'react-native'
 import { useFocusEffect, useNavigation } from '@react-navigation/native'
 import useGroup from '@hylo/hooks/useGroup'
 import useRouteParams from 'hooks/useRouteParams'
-import HyloWebView from 'components/HyloWebView'
+import HyloWebView from 'components/HyloWebView/HyloWebView'
 import { amaranth, capeCod, rhino40, rhino80, twBackground } from 'style/colors'
 
 export default function GroupSettingsWebView () {
   const navigation = useNavigation()
-  const webViewRef = useRef()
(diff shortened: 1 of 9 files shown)


thomasgwatson:
🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏🙏

apps/mobile/src/screens/GroupSettingsWebView/GroupSettingsWebView.js
```diff
@@ -1,27 +1,18 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import React, { useCallback } from 'react'
 import { ScrollView, StyleSheet, Text, TouchableOpacity, View } from 'react-native'
 import { useFocusEffect, useNavigation } from '@react-navigation/native'
 import useGroup from '@hylo/hooks/useGroup'
 import useRouteParams from 'hooks/useRouteParams'
-import HyloWebView from 'components/HyloWebView'
+import HyloWebView from 'components/HyloWebView/HyloWebView'
```
thomasgwatson, line 6: Does this matter? Importing it directly instead of via its index?

thomasgwatson (approved):
lgtm, just one little question

apps/mobile/src/screens/GroupSettingsWebView/GroupSettingsWebView.js
lorenjohnson: Good catch. It won't break anything, but isn't correct. My auto-import in Visual Studio code does this by default and sometimes forget to correct it. Fixed now.

PR #542, opened 2025-04-03
Align chat text with avatar on all media sizes; remove margin on p ta…

thomasgwatson:
…gs in chat posts

Examples at two screen sizes
Screenshot 2025-04-02 at 5 33 54 PM
Screenshot 2025-04-02 at 5 34 00 PM


tibetsprague (approved):

tibetsprague:
0 margin might be too late little for paragraphs, we do want some separation between them I think, but lets try it and keep playing

PR #546, opened 2025-04-04
538 no double votes

### LINKED ISSUE #538: Voting on a proposal in the chat room does not update the UI immediately, which also allows for multiple votes

thomasgwatson:
Closes #538

### DIFF
--- a/apps/web/src/routes/ChatRoom/ChatRoom.js
+++ b/apps/web/src/routes/ChatRoom/ChatRoom.js
@@ -417,6 +417,68 @@ export default function ChatRoom (props) {
     messageListRef.current?.data.map((item) => post.id === item.id || (post.localId && post.localId === item.localId) ? newPost : item)
   }, [group?.id])
 
+  const onAddProposalVote = useCallback((post, optionId) => {
+    const optimisticUpdate = {
+      proposalVotes: {
+        ...post.proposalVotes,
+        items: [
+          ...post.proposalVotes.items,
+          {
+            postId: post.id,
+            optionId,
+            user: currentUser
+          }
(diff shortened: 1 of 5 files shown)


apps/web/src/routes/ChatRoom/ChatRoom.js
```diff
@@ -417,6 +417,68 @@ export default function ChatRoom (props) {
     messageListRef.current?.data.map((item) => post.id === item.id || (post.localId && post.localId === item.localId) ? newPost : item)
   }, [group?.id])
 
+  const onAddProposalVote = useCallback(({ post, optionId }) => {
```
tibetsprague, line 420: this is great for now, but im realizing this is a lot of stuff we have to do for every single action that could update a post card in the chat stream. there are more and there will be more i think. so im wondering if we can generalize this and just have a function that takes the optimistically updated post data in the redux store and does a general replace of the whole post data in the chats with that 🤔 just a pondering for the future

tibetsprague (approved):

apps/web/src/routes/ChatRoom/ChatRoom.js
lorenjohnson: If there is an URQL conversion in the future for Web, I think you'll find that all of this will be handle implicitly as a function of graph cache.

thomasgwatson: Yeah, this was inspired me checking about tracks/actions going into the chat...

It'll get out of hand real fast.

As Loren said, might be good justification to push for urql here sooner rather than later

PR #574, opened 2025-04-23
Tracks

tibetsprague:


apps/backend/migrations/20250320172703_tracks.js
```diff
@@ -0,0 +1,87 @@
+exports.up = async function (knex) {
+  await knex.schema.createTable('tracks', table => {
+    table.increments().primary()
+    table.string('name').notNullable()
+    table.text('description')
+    table.text('banner_url')
+    table.string('actions_name')
+    table.text('welcome_message')
+    table.text('completion_message')
+    table.string('completion_badge_name')
```
thomasgwatson, line 10: So this isn't a direct reference to an existing group badge?

apps/backend/api/graphql/mutations/convertGraphqlData.js
```diff
@@ -1,6 +1,7 @@
 import { transform, snakeCase } from 'lodash'
 
 export default function convertGraphqlData (data) {
+  console.log('data', data)
```
thomasgwatson, line 4: I know this is a draft but just tagging this one so it will still come up in the PR history if its still around for the actual review

apps/web/src/store/actions/trackActions.js
```diff
@@ -0,0 +1,327 @@
...
+export const ENROLL_IN_TRACK_PENDING = `${MODULE_NAME}/ENROLL_IN_TRACK_PENDING`
+export const FETCH_TRACK = `${MODULE_NAME}/FETCH_TRACK`
+export const LEAVE_TRACK = `${MODULE_NAME}/LEAVE_TRACK`
+export const LEAVE_TRACK_PENDING = `${MODULE_NAME}/LEAVE_TRACK_PENDING`
+export const UPDATE_TRACK = `${MODULE_NAME}/UPDATE_TRACK`
+export const UPDATE_TRACK_PENDING = `${MODULE_NAME}/UPDATE_TRACK_PENDING`
+export const UPDATE_TRACK_ACTION_ORDER = `${MODULE_NAME}/UPDATE_TRACK_ACTION_ORDER`
+export const UPDATE_TRACK_ACTION_ORDER_PENDING = `${MODULE_NAME}/UPDATE_TRACK_ACTION_ORDER_PENDING`
+
+const CommentFieldsFragment = `
```
thomasgwatson, line 15: Could we re-use an existing fragment here? 

'@hylo/graphql/fragments/commentFieldsFragment' looks almost identical?

apps/backend/migrations/20250320172703_tracks.js
tibetsprague: you know, that would be easier and better. good call. i was just following aaron's designs

apps/web/src/store/actions/trackActions.js
tibetsprague: probably for the CommentFields yes

apps/backend/api/models/Group.js
```diff
@@ -1002,6 +1012,22 @@ module.exports = bookshelf.Model.extend(merge({
           }
         }
 
+        // Handle track case
+        if (track && track.published_at) {
+          // Only add tracks widget if it is published
+          const tracksWidget = widgets.find(w => w.get('view') === 'tracks')
+          if (tracksWidget && !tracksWidget.get('auto_added')) {
```
thomasgwatson, line 1019: There is a little trap here: If an admin adds the tracks widget to the menu before they create a track, then auto_added will be switched to true. There is then no condition where the admin visibility will ever be switched away from admin, meaning regular users will never see the tracks widget.

Probably best to separate out these checks/operations and make sure they can happen independent of one another

apps/web/src/components/PostDialog/PostDialog.jsx
```diff
@@ -12,12 +12,15 @@ const PostDialog = ({
   const navigate = useNavigate()
   const location = useLocation()
 
+  // Add state to track the container's scroll position
```
thomasgwatson, line 15: Whats all this for?

apps/web/src/routes/PostDetail/PostDetail.js
```diff
@@ -398,14 +398,14 @@ export function ActionCompletionsSection ({ currentUser, post }) {
...
-    <div className='border-border border-2 rounded-lg p-4 flex flex-row gap-2'>
+    <div className='border-border border-2 rounded-lg p-4 flex flex-row gap-2 mt-4 w-full justify-between'>
       <PeopleInfo
         people={completers}
         peopleTotal={completers.length}
         onClick={toggleCompletionResponsesDialog}
         excludePersonId={currentUser.id}
         phrases={{
-          emptyMessage: t('No one has completed this action yet'),
+          emptyMessage: t('No one has completed this - be the first!'),
```
thomasgwatson, line 408: Changing the key for a piece of translated text just means this won't find any values in the locales and will just show the key in all instances.

Please make sure the key matches a value in both the en.json and es.json for web, and that the value represents the copy that you want

apps/web/src/routes/TrackHome/TrackHome.jsx
```diff
@@ -68,7 +68,7 @@ function TrackHome () {
   const { setHeaderDetails } = useViewHeader()
   useEffect(() => {
     setHeaderDetails({
-      title: t('Track: {{trackName}}', { trackName: currentTrack?.name || 'loading...' }),
+      title: t('{{trackName}}', { trackName: currentTrack?.name || 'loading...' }),
```
thomasgwatson, line 71: Similarly here, if we are just passing the name in and it has no other copy, we should just remove the t function altogether and just directly pass the variable in

tibetsprague: yeah @brodeur needs to be trained up in the internationalization stuff

apps/web/src/components/PostDialog/PostDialog.jsx
tibetsprague: sometimes the post dialog appears in the wrong place when the outer page is scrolled down. I saw this too though I actually want to resolve it by not using the shadcn/radix PostDialog anymore and going back to our own Modal. later though

apps/backend/api/models/Group.js
tibetsprague: ooh, interesting

tibetsprague:
@thomasgwatson i'm testing on mobile and have a few more things to do: 
- [ ] Clicking on a person in the People list for a Track should take you to the profile for that person
- [ ] The "Current Action" should be called "Next Action" is it is on web, and ideally the colors would match, with the status pill having an orange border. Completed actions should have a blue border, same as web. 
- [ ] Remove the Posted In line from the Action posts, since they will always just be in the current group only right now
- [ ] When viewing an action there is a large empty header at the top of the view (see image), can we remove that?
- [ ] Add some margin above the tab bar in the track
- [ ] Clicking on the Actions tab of a Track would ideally reload it. Or maybe just clicking on the track reloads the whole thing. Right now I can't figure out how to get it to load new actions that have been added to the track without force quitting the app entirely.
- [ ] The completion results text for multi-select actions shows the selected values twice. 
- [ ] Clicking on the text box to fill out the response to a text response action is hiding the text box under the weird big empty margin at the top of the action
- [ ] Clicking on the upload attachment action doesnt show the action content at all, only the upload an attachment button, so there's no way to see the full action details

tibetsprague:
IMG_8825
IMG_8826
IMG_8827
IMG_8828

tibetsprague:
I'm not sure any of these are a blocker to deploying, but would be great to get as many done as possible before we deploy, this week!

PR #610, opened 2025-05-26
Check whether there is a groupSlug available from the matched path, a…

### LINKED ISSUE #582: Push notifications often not going to correct view still
Trying to document these closely to track this down. Sometimes they work, sometimes they go to some other group, maybe the last one I was looking at, maybe some completely other group.

thomasgwatson:
…nd use that to avoid certain navigation changes

Closes #582 (We hope...)

### DIFF
--- a/apps/mobile/src/hooks/useHandleCurrentGroup.js
+++ b/apps/mobile/src/hooks/useHandleCurrentGroup.js
@@ -15,12 +15,13 @@ import getActiveRoute from 'navigation/getActiveRoute'
 
 export function useHandleCurrentGroupSlug () {
   const { currentGroupSlug } = useCurrentGroupStore()
-  const { context, groupSlug } = useRouteParams()
+  const { context, originalLinkingPath } = useRouteParams()
   const [{ currentUser }] = useCurrentUser()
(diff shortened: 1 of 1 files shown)


apps/mobile/src/hooks/useHandleCurrentGroup.js
```diff
@@ -51,6 +52,9 @@ export function useHandleCurrentGroup () {
   const [{ currentGroup, fetching: fetchingCurrentGroup }] = useCurrentGroup()
   const loading = fetchingCurrentUser && fetchingCurrentGroup
 
+  const { context, groupSlug, originalLinkingPath } = useRouteParams()
+  const groupSlugFromPath = originalLinkingPath?.match(/\/groups\/([^\/]+)(?:\/|$)/)?.[1] ?? null
```
thomasgwatson, line 56: In my mobile tracks branch, I have created a little pathHelper in /packages to check if there are tracks related params in the originalLinkingPath. Once that is merged across, I would move these little regexes into there

tibetsprague (approved):
would love a @lorenjohnson look, but lgtm

PR #621, opened 2025-06-12
Mobile 6.1.0 into main

thomasgwatson:


thomasgwatson:
How else are we merging down to main? Should I have grabbed the release branch and just merged that to main?

tibetsprague:
ah yes, I guess ive always just done a manual direct merge into main for deploy. this works too but we would just have to coordinate. in the meantime, this one has already been merged no? so can we remove?

PR #622, opened 2025-06-12
DRAFT: 612 display join questions

### LINKED ISSUE #612: Display members’ answers to join questions to stewards
To implement: 
- [ ] The Join Requests tab in Settings will have Pending Requests and a link to the member directory for existing members and their join questions
- [ ] In the members view, for admins:
- [ ] Toggle for Show Join questions. Put it next to Sort by Name
- [ ] When true, display the responses to the join questions
- [ ] Sort toggle moves with you as you scroll down

thomasgwatson:
Closes #612


apps/web/src/routes/Members/Members.js
```diff
@@ -120,6 +126,16 @@ function Members (props) {
...
             }))}
           />
+          {canSeeJoinAnswers && (
+            <div className='flex items-center gap-2'>
+              <SwitchStyled
+                checked={showAnswers}
+                onChange={() => setShowAnswers(!showAnswers)}
+                backgroundColor={showAnswers ? '#0DC39F' : '#8B96A4'}
+              />
+              <span className='text-sm font-medium text-foreground/80'>{t('Show Answers')}</span>
```
tibetsprague, line 136: this should probably say something like Show Join Question Answers, or Show Join Answers

thomasgwatson: Yeah, its just such a small space.... having that much text will make that button dominate

PR #625, opened 2025-06-14
585 send event rsvp invite as attachment

### LINKED ISSUE #585: Send an email with a calendar invite attachment when someone RSVPs going or interested to an Event

KevinTriplett:
closes #585 

* Successfully send iCal file that is accepted by Apple Calendar and Google Calendar.
* [ ] Needs testing with Outlook calendar.
* Added needed mobile dev file updates
* Set cursor to pointer for RSVP dropdown list
* [ ] The SendWithUs RSVP email template needs some TLC -- the format is a draft. For example, the subject line is missing the actual response.
* Adds the [iCalendar library](https://sebbo2002.github.io/ical-generator/develop/reference/index.html) 
* The "group_name" is a join of all the groups that this event was posted to -- I'm not sure what the current practice is, to take the first group or what. There seems no indication on the backend of where this event was viewed and RSVP'ed.


apps/backend/api/graphql/mutations/event.js
```diff
@@ -17,11 +18,42 @@ export async function respondToEvent (userId, eventId, response) {
...
       response
     })
   }
+
+  if (response === EventInvitation.RESPONSE.YES ||
+    response === EventInvitation.RESPONSE.INTERESTED
+  ) {
+    const iCalendar = ical()
+    const iCalData = await event.getCalData(userId)
+    iCalendar.createEvent(iCalData)
```
tibetsprague, line 32: so the only thing im now realizing is that this will slow down the mutation a bit, if it still seems very fast then this is fine, but the other option is to move all this generation into the Email.sendEventRsvpEmail. i'll leave that decision to you.

tibetsprague (approved):
whoo hoo! i can fix up the template. only one comment here to consider

Issue #556, opened 2025-04-15
Posts cannot be tagged #home


thomasgwatson:
So it makes it hard to get a post into #home chatrooms when the TO: field is being weird

![Image](https://github.com/user-attachments/assets/38808f51-789a-4daf-94b7-e1bc698f7976)

thomasgwatson:
@tibetsprague I don't think I can actually fix this one, because all my local db groups have #home as one of their suggested topics. What do we need to change about group_tags to get all the home tags to be suggested? change `is_default` = `true`?

PR #628, opened 2025-06-16
Load #topics much faster in posts

tibetsprague:
Make the topic selector much faster
    
    Also don't show duplicate create option in the add Chat Room selector when typing in an existing topic in the group.
    
    Fixes https://github.com/Hylozoic/hylo/issues/556


thomasgwatson (approved):
lgtm

PR #620, opened 2025-06-12
Mobile improvements

brodeur:
A number of mobile improvements, including signup flow, improved local testing abilities, making things display better at smaller screen sizes throughout

PR #637, opened 2025-06-19
More mobile improvements

tibetsprague:
Fix z-index of stream view controls


tibetsprague:
see comments here for more work to do https://github.com/Hylozoic/hylo/pull/620

thomasgwatson (approved):
lgtm

tibetsprague:
this is waiting on @brodeur to do this additional tasks from the other PR linked

brodeur:
OK I've addressed the other mobile issues you discovered, and then some @tibetsprague

apps/web/src/components/Dropdown/Dropdown.js
```diff
@@ -4,24 +4,41 @@ import { cn } from 'util/index'
...
 import { position } from 'util/scrolling'
 import Icon from 'components/Icon'
+import { useDropdown } from 'contexts/DropdownContext'
 import classes from './Dropdown.module.scss'
 
-const Dropdown = ({ children, className, triangle, items, toggleChildren, alignRight, menuAbove, noOverflow }) => {
+const Dropdown = ({ children, className, triangle, items, toggleChildren, alignRight, menuAbove, noOverflow, id }) => {
   const [active, setActive] = useState(false)
   const parentRef = useRef(null)
+  const { activeDropdownId, openDropdown, closeAllDropdowns } = useDropdown()
```
tibetsprague, line 13: dang, should have mentioned that we are also trying to move away from our Dropdown to the shadcn/radix components/ui/select, which I think handles all this already. but would need to be tested more. generally we are trying to move as much as possible to the shadcn components which are tailwind wrappers around Radix components

tibetsprague: but for now we can roll with this

PR #644, opened 2025-06-24
Color scheme

brodeur:
Peep dis y'all. We need to present a few options to our community, but I want to try some out with you all. This is just changing a few variables. It makes me think theming wouldn't actually be so hard.

### DIFF
--- a/apps/web/src/components/PostSelector/PostSelector.js
+++ b/apps/web/src/components/PostSelector/PostSelector.js
@@ -159,7 +159,7 @@ export default function PostSelector ({ collection, draggable, group, onRemovePo
           <div className='relative'>
             <div>
               <input
-                className='bg-input/60 focus:bg-input/100 rounded-lg text-foreground placeholder-foreground/40 w-full py-1 px-2 transition-all outline-none focus:outline-focus focus:outline-2'
+                className='bg-input/60 focus:bg-input/100 rounded-lg text-foreground placeholder-foreground/40 w-full py-2 px-2 transition-all outline-none focus:outline-focus focus:outline-2'
                 ref={searchBoxRef}
                 type='text'
                 placeholder={t('Search for posts')}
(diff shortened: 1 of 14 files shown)


apps/web/src/routes/UserSettings/UserSettings.js
```diff
@@ -83,10 +84,15 @@ const UserSettings = () => {
       component: <ManageInvitesTab currentUser={currentUser} />
     },
     {
-      name: t('Locale'),
-      path: 'locale',
+      name: t('Language'),
```
tibetsprague, line 87: actually im now wondering if we have a General Settings tab that has language and appearance related stuff. otherwise we have two tabs that each only have 1 or 2 settings on them right now

apps/web/src/components/ThemeSelector/ThemeSelector.js
```diff
@@ -0,0 +1,83 @@
...
+    availableThemes
+  } = useTheme()
+
+  return (
+    <div className={cn('flex flex-col gap-4', className)}>
+      <div className='flex items-center justify-between'>
+        <label className='text-sm font-medium'>{t('Color Scheme')}</label>
+        <div className='flex items-center gap-2 rounded-lg border p-1'>
+          <button
+            onClick={() => setColorScheme('light')}
```
tibetsprague, line 32: we should have the option here to choose system, so it continues to change based on the system's dark or light mode choice

tibetsprague: and that should continue to be the default for folks who have not chosen otherwise

tibetsprague (changes requested):
i think for now we roll this out without the color theme choices, just the dark/light/system chooser. and the new color updates. though maybe we run those by the team?

apps/web/src/routes/UserSettings/UserSettings.js
brodeur: Let's discuss

tibetsprague:
All that being said its pretty amazing that these work as well as they do! We are definitely close to ready.

tibetsprague:
oh one other comment, i still think the new Dark mode does not have enough contrast. The icons in the global nav blend into the background. As do post cards in the chat, for example

tibetsprague:
In lightmode, it looks like the emoji hover is in dark mode and is hard to read
Screenshot 2025-07-29 at 9 56 17 PM

tibetsprague:
i do kind of feel that the nav menu is now too dark. looks too much like dark mode. that could just be me though. Also the divider between pinned is impossible to see 
Screenshot 2025-07-29 at 9 57 21 PM

tibetsprague:
this set of colors all together looks ugly to me. maybe its just the request still needed box thats off? also whats up with the square corners at the top of the post content? thats on production too. 

Screenshot 2025-07-29 at 10 01 51 PM

tibetsprague:
i liked it better before when the modal background was dark, not light
Screenshot 2025-07-29 at 10 03 46 PM

tibetsprague:
also, i have to be honest. i like the current light mode background colors better than the new one 😬 maybe im just use to them but the new ones look now a little too green.

tibetsprague:
snow is ok, except the notification color doesnt fit i dont think 
Screenshot 2025-07-29 at 10 07 17 PM

tibetsprague:
i also dont like jungle. honestly the only new one i like ok is Ocean 😬 sorry im a harsh critic here! I would not feel great about releasing these as our first new themes. but lets defintely get some other opinions!

tibetsprague:
Screenshot 2025-10-07 at 5 05 56 PM
placeholder text in snow/dark is impossible to read

brodeur:
OK I think I addressed all of these, except I'm not sure if I fixed the placeholder text one. Let me know what you think!

PR #645, opened 2025-06-24
My home improvements

brodeur:


apps/web/src/routes/UserSettings/NotificationSettingsTab/NotificationSettingsTab.js
```diff
@@ -132,88 +127,95 @@ function NotificationSettingsTab ({
...
-                <SelectItem value='never'>{t('Never')}</SelectItem>
-                <SelectItem value='mixed' disabled>{t('~ Mixed ~')}</SelectItem>
-              </SelectContent>
-            </Select>
+        </div>
+        <div className='text-sm text-foreground/50 mb-2 mt-4'>{t('Default group notifications')}</div>
+        <div className='bg-card/100 rounded-lg p-4 shadow-lg mb-4'>
+          <div>
+            <div className='flex items-center text-sm text-foreground/80 mb-2'>
+              <span>{t('These settings apply to all groups you are a member of. Any group-specific settings will override these defaults for that group.')}</span>
```
tibetsprague, line 151: this is not quite accurate. the way these toggles work is turning it on turns on the setting for all the groups, turning it off turns off the setting for all the groups. if all the groups have this setting on the toggle will show as on, if even one group has it turned off then the toggle will show as off here. that could be a little confusing, ideally we would show it in some sort of half/mixed state if some groups have it on and others off. but there's not a default/override thing happening

apps/web/src/components/GroupButton/GroupButton.js
```diff
@@ -1,15 +1,21 @@
 import React from 'react'
 import { groupUrl } from 'util/navigation'
 import { DEFAULT_AVATAR } from 'store/models/Group'
-import { Link } from 'react-router-dom'
-import Button from 'components/Button'
+import { useNavigate } from 'react-router-dom'
+import Button from 'components/ui/Button'
```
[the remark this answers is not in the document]
tibetsprague: fixed

apps/web/src/routes/UserSettings/NotificationSettingsTab/NotificationSettingsTab.js
brodeur: Ok, I updated the language to try to reflect what I'm understanding here -- what do you think?

brodeur:
Ok I made a bunch more progress on this branch, let me know what you think

Issue #662, opened 2025-06-30
Mobile: post detail exit navigates to known path, instead of relying on history


thomasgwatson:
At the moment, a chat post with a comment, or a specific post view, will open the Post Details modal. When the user is done with that modal, the exit function just shows whatever remains on the stack. That is fine for when this navigation happens in the app, since the stack and the path will align. 

However, when the use arrives at this view from a notification, there are many times when the path from the notif and the stack history do not align.

So, we should not rely on the stack history in these instance and actively navigate when the user is done with that detail view

thomasgwatson:
Checked this and its defacto solved by other changes

PR #663, opened 2025-06-30
661 chat paths happier

### LINKED ISSUE #661: Mobile: Comments are actually respected by mobile
Actually highlight comments if they are part of the URL

~~An example path `/groups/hylo-stewards/chat/general/post/90088`~~

~~In mobile currently, we have separate match records for the main view (`/groups/hylo-stewards/chat/general` in this instance) and the detail view (`/post/90088`). So the path resolves to the detail view, which is not what we want for chat posts.~~

~~I'm not quite sure how to avoid this in the current mobile setup. The path that is matching it~~  `':unmatchedBasePath(.*)/post/:id'` ~~but there is also a~~ '/:context(groups)/:groupSlug/chat/:topicName/post/:id' ~~above it that should match it as well.~~

thomasgwatson:
Closes #661  #662

Doing some mobile clean up around behavior related to the outcomes of push notifications and how it impacts navigation


apps/mobile/src/navigation/headers/ModalHeader.js
```diff
@@ -30,6 +32,23 @@ export default function ModalHeader ({
   ...otherProps
 }) {
   const confirmAlert = useConfirmAlert()
+  const openURL = useOpenURL()
```
tibetsprague, line 35: add comments here please

apps/mobile/src/navigation/headers/ModalHeader.js
```diff
@@ -49,9 +68,8 @@ export default function ModalHeader ({
       fontFamily: 'Circular-Bold'
     },
     headerLeft: headerLeft || options.headerLeft || (props => {
-      // get go back function from navigation
       const headerLeftOnPress = options.headerLeftOnPress ||
-        providedHeaderLeftOnPress ||
+        providedHeaderLeftOnPress || respectOriginalPath ||
```
tibetsprague, line 72: also comments here

tibetsprague (approved):

Issue #417, opened 2025-03-15
HyloHTML for mobile: Verify anchor tags work as expected and that other tags work as well


thomasgwatson:
For example: building hylo description. Easiest example to find might be adding a fancy html description to a group and then displaying the ABOUT widget in the menu

![Image](https://github.com/user-attachments/assets/8fc46d45-cc15-42d8-99cf-131b1402bf0e)

lorenjohnson:
@thomasgwatson sorry about this, but I'm still not clear what the open issue or question is here? I think in the screenshot you provided it seems the embedded links were not links or auto-converted to be so. The current HyloHTML component does handle links, and formatting. Please be more specific about what you tried, what you're trying to achieve, and what didn't work as expected.

thomasgwatson:
On mobile
![Image](https://github.com/user-attachments/assets/4410b063-1856-4013-877d-b586a3969139)

On web but notice the big differences between the about widget and the about menu.
![Image](https://github.com/user-attachments/assets/2ee826a8-dca3-40b6-a242-6dee5a24067e)

So whatever is happening in the About Page needs to be ported to both web and mobile menus

PR #672, opened 2025-07-01
568 welcome page refresh

thomasgwatson:
Inspired by #668 and #417


apps/mobile/src/screens/GroupWelcome/GroupWelcome.js
```diff
@@ -202,7 +203,7 @@ function LandingBodyContent ({ description, purpose, welcomePage }) {
     <>
       {welcomePage &&
         <View style={{ flexDirection: 'column', justifyContent: 'center' }}>
-          <HyloHTML html={welcomePage} />
+          <HyloHTML html={TextHelpers.markdown(welcomePage)} />
```
tibetsprague, line 206: we don't support markdown on the web right now. seems a little confusing to have both markdown and HTML support no? i mean i think its fine but just wanted to hear your reasoning.

tibetsprague (approved):

PR #681, opened 2025-07-03
629 send rsvp email only on important change

### LINKED ISSUE #629: Don't resend event RSVP email on every response change
No need to send email if response goes from Going to Interested and vice versa.

Okay to send email if response goes from either of them to Not Going and vice versa.

KevinTriplett:
closes #629 

* Send rsvp (invite.ics) email when user indicates going
* Send when user changes from going to not going
* Send when user changes from not going to going
* Send cancelation invite.ics when not going

Going is "GOING" and "INTERESTED"


apps/backend/api/graphql/mutations/event.js
```diff
@@ -13,30 +18,35 @@ export async function respondToEvent (userId, eventId, response) {
     throw new GraphQLError('Event not found')
   }
 
-  const eventInvitation = await EventInvitation.find({ userId, eventId })
+  let eventInvitation = await EventInvitation.find({ userId, eventId })
+  const sendEmail = (!eventInvitation && going(response)) ||
```
tibetsprague, line 22: seems like this could error out if !eventInvitation and !going(response), because the event invitation would not exist, but I guess that should never happen? would be safest to add question marks like !going(eventInvitation?.get('response') just in case

apps/backend/api/models/Post.js
```diff
@@ -348,20 +348,30 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
   // for event objects, for use in icalendar
   // must eager load the user relation
-  getCalEventData: function (forUserId) {
+  getCalEventData: function (forUserId, eventInvitation) {
     const user = this.relations.user
+
+    const notGoing = eventInvitation.get('response') === EventInvitation.RESPONSE.NO
+    const status = notGoing ? 'CANCELLED' : 'CONFIRMED'
+    const method = notGoing ? 'CANCEL' : 'REQUEST'
+    const ical_sequence = (eventInvitation.get('ical_sequence') || 0) + 1
```
tibetsprague, line 357: seems a little odd to add 1 here as well, why not just use the sequence that's in the database? that makes more sense to me. adding +1 in two places feels brittle to me

tibetsprague (changes requested):
great! a couple small things but otherwise looks good

tibetsprague (approved):

Issue #608, opened 2025-05-22
Cannot navigate away from chat post


tibetsprague:
I clicked on a link to a chat post and got to this screen

![Image](https://github.com/user-attachments/assets/28dc0d60-0d63-435a-bdca-52ed2507b6df)

The x button is impossible to click, and there is no menu, so there is no way to get off this screen. Even the link to the Hylo General Circle doesn't work.

thomasgwatson:
Do you know what context the initial link was in? I'm guessing you clicked it in the chatroom webview. And that we aren't actively handling `/chat` paths, so it navigates to a webview version of that path and things get weird

thomasgwatson:
I have made some other related corrections to paths for posts and chats in the tracks branch, so that might have snagged this already. Let's see if we can recreate after we gets Tracks out

thomasgwatson:
Just wanting to keep tabs on this; Let me know if you see it again

thomasgwatson:
so, in less than exciting news;

Getting modals to work on cold boot apparently requires a rewrite of how we handle deep-linking

Claude suggests this is how long it would take

Initial Implementation: 2-3 days
Testing & Refinement: 1-2 days
Total: ~5 days

All this just to try to have the core content load first, instead of grinding the app through rendering/loading everything else first...

Probably better to flip the Post Details view to be a regular screen instead of a modal

thomasgwatson:
There are some changes to how React Navigation has setup `onReady` and `navigationRef.isReady()`, in the latest major version (7), so it might be time to push for that

Issue #690, opened 2025-07-06
Mobile app: Clicking on a link to a group i am not a member of leads to blank screen


tibetsprague:
Someone posted a link to a Hylo group on a signal thread, i clicked on it and it takes me to the app but to a blank screen (the with navigation showing at the bottom). Bummer. This should take me to the group details/landing page where i can join or request to join. The group is publicly visible as i tried it on the web and it worked.

hylobot:
Hmm. Not for me. It happens even when the app was just opened. On Jul 6, 2025, at 2:20 PM, 'Tom Watson' via Developers ***@***.***> wrote:﻿thomasgwatson left a comment (Hylozoic/hylo#690)
Ok, this turned out to be another cold-boot navigation issue. Works totally fine if the app is already running but shows nothing on cold-boot.

Issue #687, opened 2025-07-04
My Interactions navigation bounces to My Posts from the menu


thomasgwatson:
When navigating to the My Interactions view from the context menu, the My Posts view is also loaded. But this only happens the first time you try to make this navigation.

Something is funky

Issue #686, opened 2025-07-04
Mobile: header.goBack functionality is now sometimes inserting something in the navigation stack


thomasgwatson:
Probably related to the navigation #662 shenans

PR #695, opened 2025-07-09
Better handle when to go to homeWidget on navigation actions

thomasgwatson:
Closes:
#608
#686  
#687 
#690 

By-product of this PR: the global nav and the context menu are now available as regular screens; this provides a user-friendly fallback in specific navigation edge-cases.


apps/mobile/src/hooks/useHandleCurrentGroup.js
```diff
@@ -10,19 +10,20 @@ import mixpanel from 'services/mixpanel'
...
   const changeToGroup = useChangeToGroup()
-  const groupSlugFromPath = originalLinkingPath?.match(/\/groups\/([^\/]+)(?:\/|$)/)?.[1] ?? null
+  const pathMatches = originalLinkingPath?.match(/\/groups\/([^\/]+)(.*$)/)
+  const groupSlugFromPath = pathMatches?.[1] ?? null
+  const pathAfterMatch = pathMatches?.[2] ?? null
 
   useEffect(() => {
     if (currentUser && !currentGroupSlug && !groupSlugFromPath) {
-      changeToGroup(getLastViewedGroupSlug(currentUser))
+      changeToGroup(getLastViewedGroupSlug(currentUser)) // tempting to switch this to NoContextFallbackScreen
```
tibetsprague, line 26: what would this mean/do?

apps/mobile/src/hooks/useHandleCurrentGroup.js
```diff
@@ -66,13 +67,20 @@ export function useHandleCurrentGroup () {
...
+      } else if (currentGroup?.homeWidget && navigateHome) {
         setNavigateHome(false)
         // Only "replace" current HomeNavigator stack if there are mounted screens,
         // otherwise begins loading the default HomeNavigator screen then replaces
-        // it the homWidget which  creates tow navigation animations
-        const hasMountedScreens = !!navigationState?.routes[0].params
-        openURL(widgetUrl({ widget: currentGroup.homeWidget, groupSlug: currentGroup.slug }), { replace: hasMountedScreens })
+        // it the homeWidget which creates two navigation animations
+        const hasMountedScreens = !!navigation.getState()?.routes[0].params
+        if (currentGroup?.slug === 'my') {
```
tibetsprague, line 76: add a comment here to explain

apps/mobile/src/screens/NoContextFallbackScreen/NoContextFallbackScreen.js
```diff
@@ -0,0 +1,27 @@
...
+import { View } from 'react-native'
+import useCurrentGroup from '@hylo/hooks/useCurrentGroup'
+import ContextSwitchMenu from 'components/ContextSwitchMenu'
+import LoadingScreen from 'screens/LoadingScreen'
+
+export default function NoContextFallbackScreen () {
+  const [{ currentGroup, fetching }] = useCurrentGroup()
+
+  // If we're fetching or we have a currentGroup, show loading
+  // if (fetching || (currentGroup && currentGroup.slug)) {
```
tibetsprague, line 11: remove

apps/mobile/src/screens/NoContextFallbackScreen/NoContextFallbackScreen.js
```diff
@@ -0,0 +1,27 @@
...
+    if (fetching || currentGroup?.slug) {
+    return <LoadingScreen />
+  }
+
+  // Only show the context menu if we truly have no group context
+  return (
+    !currentGroup?.slug && (
```
tibetsprague, line 18: what happens if we do have a current group context here?

tibetsprague (approved):
some requests for comments and clarification but otherwise looks good

apps/mobile/src/hooks/useHandleCurrentGroup.js
thomasgwatson: - From cold-boot, the app will have no currentGroup set.
- It must open something
- So setting a group (in this case, the `lastViewedGroup`) effectively triggers side-effects to open the group at its home widget

There are a lot of 'depends on side-effects that aren't obvious" things happening for mobile navigation, a rich and bountiful terrain for bugs to thrive and persist. Some of those side-effects are coming from platform/framework behaviour and some from us having to layer them in based on how our app works

The NoContextFallback would instead not make the assumption that you want to see the last group but instead opens a (currently pretty ugly in full-screen) view of the global nav menu, but as a screen, not in the drawer. So it invites them to choose a context

apps/mobile/src/screens/NoContextFallbackScreen/NoContextFallbackScreen.js
thomasgwatson: Side-effects take you elsewhere

PR #698, opened 2025-07-10
Add ability to pin groups to the top of global nav

tibetsprague:
And drag and drop reorder them in the menu

Fixes https://github.com/Hylozoic/hylo/issues/125

TODO: work on mobile support


apps/backend/api/graphql/mutations/membership.js
```diff
@@ -1,9 +1,11 @@
+/* global GroupMembership, GroupJoinQuestionAnswer, Queue, bookshelf */
```
thomasgwatson, line 1: ?? still want this?

apps/web/src/components/ui/context-menu.jsx
```diff
@@ -0,0 +1,167 @@
+import * as React from 'react'
+import * as ContextMenuPrimitive from '@radix-ui/react-context-menu'
+import { Check, ChevronRight, Circle } from 'lucide-react'
+
+import { cn } from '@/lib/utils'
+
+const ContextMenu = ContextMenuPrimitive.Root
```
thomasgwatson, line 7: KInda hate this is called contextmenu, for the sake of new developers joining the project. Is it worth getting cursor to relabel all the related entities for this component?

apps/backend/api/graphql/mutations/membership.js
tibetsprague: this is only needed for linting when looking at the code from the root of the monorepo. so I think its worth leaving in?

I did this from the mono repo because I had AI basically do the whole first pass, backend and front-end at once (I started it a bit to give it some help)! It did a pretty good job and I mostly vibe coded the rest, only a few tweaks needed.

apps/web/src/components/ui/context-menu.jsx
tibetsprague: yeah fair. RightClickMenu? PopupMenu?

apps/web/src/store/reducers/ormReducer/index.js
```diff
@@ -776,6 +777,88 @@ export default function ormReducer (state = orm.getEmptyState(), action) {
       break
     }
 
+    case UPDATE_MEMBERSHIP_NAV_ORDER_PENDING: {
+      me = Me.first()
+      membership = Membership.safeGet({ group: meta.groupId, person: me.id })
+
+      if (!membership) break
+
+      const newNavOrder = meta.navOrder
```
thomasgwatson, line 786: Would be nice to bundle this into a platform agnostic function that can live in packages/shared. Otherwise we'll be repeating the operation in three spots

Reminds me that I should move reorderTree over to shared

thomasgwatson (approved):
lgtm

apps/web/src/components/ui/context-menu.jsx
tibetsprague: it is specifically made to be a right-click menu and yes it automatically works on mobile with a long press, which is nice.

apps/web/src/store/reducers/ormReducer/index.js
thomasgwatson: I went through that process with context widgets; separating out the logic from the platform/environment specific side-effects. Overall this logic should be simpler because its just a flat list ordering... but if we hit bugs a bunch in this, it might be worth coming back to it and pulling the logic out into a directly testable function

thomasgwatson:
What do you wanna do about the mobile part of this? Was there a specific ticket/issue already?

PR #706, opened 2025-07-15
Send image url to OneSignal for image use in push notifications

thomasgwatson:
I'm not super invested in this one; happy to try and see if our existing images will work as references for the push notifications. Will depend on whether those images are formatted/clipped to the specifications that OneSignal desires, and whether the advice the OneSignal chatbot provided is actually accurate.

### DIFF
--- a/apps/backend/api/models/Notification.js
+++ b/apps/backend/api/models/Notification.js
@@ -8,6 +8,44 @@ import { en } from '../../lib/i18n/en'
 import { es } from '../../lib/i18n/es'
 const locales = { en, es }
 
+const DEFAULT_USER_AVATAR = 'https://www.gravatar.com/avatar'
+const DEFAULT_GROUP_AVATAR = '/default-group-avatar.svg'
+
+async function getIconUrl (notification) {
+  const { activity } = notification.relations
+  const reasons = activity.get('meta').reasons
+  const reason = Notification.priorityReason(reasons)
+  const groupRelatedReasons = [
+    'joinRequest',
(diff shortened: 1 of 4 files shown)


apps/backend/api/models/Notification.js
```diff
@@ -8,6 +8,44 @@ import { en } from '../../lib/i18n/en'
...
+  const groupRelatedReasons = [
+    'joinRequest',
+    'approvedJoinRequest',
+    'announcement',
+    'groupChildGroupInvite',
+    'groupChildGroupInviteAccepted',
+    'groupParentGroupJoinRequest',
+    'groupParentGroupJoinRequestAccepted',
+    'newPost',
+    'memberJoinedGroup'
```
tibetsprague, line 27: I think we should add mention, chat, comment, trackCompleted, trackEnrollment. Those all are within groups

tibetsprague: I actually dont think we should ever show a user's avatar, except maybe for a DM with just that person? but that might be worth discussing

tibetsprague (changes requested):
does OneSignal do the adding of the Hylo icon to the bottom of it?

apps/backend/api/models/Notification.js
thomasgwatson: Is the mini-app icon over other icon a common pattern? Is that an iOS thing?

thomasgwatson:
We should just redo this work when we have time to focus on it

PR #725, opened 2025-07-28
Subscriptions fix

thomasgwatson:
Ok, after much debugging, it seemed that a silent connection limit was causing endless fetching on Android.

As a consequence, I have consolidated several subscriptions into a unified one. So we will have one main subscription, which will sync several different entities, and also a few context specific subscriptions, that only subscribe when specific screens are opened.


tibetsprague (approved):
Bravo!

PR #735, opened 2025-07-30
Update react native build libraries also to 0.76.9

tibetsprague:
Also give us more memory to work with in building.

I forgot to update all the related libraries. 
This is working for me


tibetsprague:
also pushed one tiny commit that speeds up the stream by many seconds with no loss of functionality! Used Reactotron to track it down

PR #762, opened 2025-08-13
Send rsvp emails on event updates

### LINKED ISSUE #689: When event details change send an update email to RSVPed people
We should let all RSVPed people know when the event time changes for sure. Also maybe if the title or description changes? not sure about that. 

Also send an updated calendar invite with the changes.

KevinTriplett:
closes #689 

Also sends event canceled emails to RSVP


KevinTriplett:
I don't know why the "fix Maximum update depth exceeded" commit is doing in this...!

apps/backend/api/graphql/mutations/post.js
```diff
@@ -30,9 +30,17 @@ export function deletePost (userId, postId) {
...
+    }).then(() => {
       return Post.deactivate(postId)
-    })
-    .then(() => ({ success: true }))
+    }).then(() => ({ success: true }))
+}
+
+export async function cancelEvent (post) {
+  await post.load('eventInvitations')
+  const eventInvitations = await post.relations.eventInvitations.fetch()
```
tibetsprague, line 41: this is weird, you shouldn't have to do a .load and a .fetch, isn't that loading eventInvitations twice? why not just do await post.eventInvitations().fetch()?

tibetsprague: also is there a reason you load these ids here and pass them into sendEventCancelRsvps? seems like sendEventCancelRsvps could just load them from the post itself and that would make more sense to me

apps/backend/api/models/Post.js
```diff
@@ -1211,6 +1256,76 @@ module.exports = bookshelf.Model.extend(Object.assign({
...
+    eventInvitations.forEach(eventInvitation => {
+      if (!eventInvitation.notGoing()) {
+        Post.sendEventRsvp({eventId: postId, eventInvitationId: eventInvitation.id, eventChanges})
+      }
+    })
+  },
+
+  sendEventCancelRsvps: async function ({ postId, eventInvitationIds }) {
+    const post = await Post.where({ id: postId }).fetch() // post is likely deactive, so fetch manuely
+    for (const eventInvitationId of eventInvitationIds) {
```
tibetsprague, line 1274: I think we should load these from the post here, not pass them in. what if they changed in the time between when the job was queued and called?

apps/web/src/routes/AuthLayoutRouter/components/GlobalNav/GlobalNav.jsx
```diff
@@ -133,9 +133,10 @@ export default function GlobalNav (props) {
     return () => {
       if (menuTimeoutId) {
         clearTimeout(menuTimeoutId)
+        setMenuTimeoutId(null)
```
tibetsprague, line 136: lets remove this from this PR, I still havent reproduced this.

tibetsprague (approved):

PR #763, opened 2025-08-14
Offline persistence of cache

thomasgwatson:
This seems like its a pretty straight-forward win. Seems to save about 4-5 seconds off cold-boot for the simulator/emulator.

Note, that this doesn't use the cache for the first load of currentUser, so there is still some separate work that will need to be tested there.


tibetsprague (approved):
Wow that’s it??

PR #773, opened 2025-08-15
Realigning mobile colors to our themes

thomasgwatson:
@tibetsprague starting to align mobile colors to light mode/theme

This is my approach to aligning mobile color/themes with web.

1) global css is source of truth (DONE)
- Nativewind uses directly
2) Run script to generate javascript variables for each color, mapped semantically based on the theme (DONE)
- import those colors and use in place of prior color system
3) build out visual mobile color alignment (DONE)
4) Tweak colors (cursor AI is colorblind; can't tell colors apart and just guesses) (DONE)
- Screenshot below

Next steps:
- go through the use of each mobile color and replace with the theme semantic color
- then at least we'll have a consistent light mode.
- Then I can go on with figuring out dark mode, since a change in mode would then notionally impact ALL COLORS, instead of just those connected to the nativewing

Screenshot 2025-08-15 at 5 11 22 PM

### DIFF
--- a/apps/backend/api/graphql/mutations/event.js
+++ b/apps/backend/api/graphql/mutations/event.js
@@ -1,6 +1,5 @@
 import { GraphQLError } from 'graphql'
 import { values, includes } from 'lodash/fp'
-import { DateTimeHelpers } from '@hylo/shared'
 
 export async function respondToEvent (userId, eventId, response) {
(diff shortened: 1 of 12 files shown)


thomasgwatson:
This is legacy now that we are using more web view screens in mobile

PR #800, opened 2025-08-20
Fix growing comment box android + more chat improvements

tibetsprague:
Fix map going below the bottom of the screen on mobile
Update chat posts immediately when the post is edited, commented or reacted on in the post editor
Fix extra scroll bar when a chat room has no posts
Make sure New Posts text always appears with new posts line
Don't double update last read post when creating a new chat


thomasgwatson (approved):
lgtm

PR #823, opened 2025-08-21
Ensure group is included in notif paths

### LINKED ISSUE #822: Clicking on post or comment from bell menu not taking me to correct group
The correct post pops up, but I stay in the current group

thomasgwatson:
closes #822

### DIFF
--- a/apps/mobile/src/screens/NotificationsList/NotificationsList.store.js
+++ b/apps/mobile/src/screens/NotificationsList/NotificationsList.store.js
@@ -113,7 +113,7 @@ export function refineActivity ({ action, actor, comment, group, post, track, me
         body: `${t('wrote:')} "${truncateHTML(post.details)}"`,
         header: t('New Post in'),
         objectName: topic,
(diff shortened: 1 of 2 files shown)


apps/mobile/src/screens/NotificationsList/NotificationsList.store.js
```diff
@@ -113,7 +113,7 @@ export function refineActivity ({ action, actor, comment, group, post, track, me
         body: `${t('wrote:')} "${truncateHTML(post.details)}"`,
         header: t('New Post in'),
         objectName: topic,
-        onPress: () => openURL(`/post/${post.id}/chat/${topic}?postId=${post.id}`)
+        onPress: () => openURL(`/groups/${group?.slug}/post/${post.id}/chat/${topic}?postId=${post.id}`)
```
tibetsprague, line 116: uh this is not right. you dont have the post id before the chat topic

tibetsprague (changes requested):
huh, so how was it working before? maybe it was opening the post in a modal? so then when you closed it would go back to the group you were previously on? or was this always broken?

PR #829, opened 2025-08-22
Correct chat context widgets type

tibetsprague:


tibetsprague:
need this actually on dev for the guy who makes our message list library to test stuff and help us track down a bug

PR #864, opened 2025-09-01
Dump cache on app crash; Add more null checks to data in threadcards;…

thomasgwatson:
… add dev only diag for urql and thread cards

### DIFF
--- a/apps/mobile/src/components/ThreadCard/ThreadCard.js
+++ b/apps/mobile/src/components/ThreadCard/ThreadCard.js
@@ -5,29 +5,36 @@ import { TextHelpers } from '@hylo/shared'
 import Avatar from 'components/Avatar'
 import { useTranslation } from 'react-i18next'
 import { rhino30, limedSpruce, nevada, rhino60, rhino, rhino10, persimmon, twBackground } from '@hylo/presenters/colors'
+import ThreadDataDiagnostic from './ThreadDataDiagnostic'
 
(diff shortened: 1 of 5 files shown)


apps/mobile/src/components/ThreadCard/ThreadCard.js
```diff
@@ -38,42 +45,60 @@ export default function ThreadCard ({ message, currentUser, onPress, participant
...
-export function ThreadAvatars ({ avatarUrls }) {
-  const count = avatarUrls.length
+export function ThreadAvatars ({ avatarUrls = [] }) {
+  const validAvatarUrls = avatarUrls.filter(Boolean)
+  const count = validAvatarUrls.length
+  
+  if (count === 0) {
+    return (
+      <View style={styles.threadAvatars}>
+        <Avatar avatarUrl={null} style={styles.firstThreadAvatar} />
```
tibetsprague, line 84: avatarUrl={null}? why would we even have this?

tibetsprague (approved):
I like the dumping cache stuff. all the thread diagnostics im not a huge fan of leaving around over time, so maybe once we're confident its not broken anymore we remove most or all of it?

apps/mobile/src/components/ThreadCard/ThreadCard.js
thomasgwatson: So Avatar component knows to use the default avatar. This is for the edge-case of when there are no valid avatar urls

PR #886, opened 2025-09-03
Create group fix and cleanup

### LINKED ISSUE #880: Creating a Group on Mobile App
After adding content to all of the fields to create a new group and then clicking the 'Let's Do This' button I was presented with a message saying 'Oops. Something went wrong' and an image of our friendly axolotl mascot

https://hylo.canny.io/admin/board/bugs-feature-requests/p/creating-a-group-on-ios

thomasgwatson:
closes #880 

Initial problem was that the return value from the mutation didn't include the group property on the membership, so yes... a partial result gets returned and a null value breaks things.


apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -33,10 +33,13 @@ export default function ContextSwitchMenu ({ isExpanded, setIsExpanded, fullView
   // Sort pinned by navOrder, unpinned by group name
   const sortedPinnedGroups = pinnedMemberships
     .sort((a, b) => a.navOrder - b.navOrder)
-    .map(m => GroupPresenter(m.group))
+    .map(m => m.group ? GroupPresenter(m.group) : null)
+    .filter(Boolean) // Remove null entries
```
tibetsprague, line 37: honestly i dont love having all these safety checks. yes it may prevent a crash but then it would lead to a more subtle bug we wouldnt catch, of groups just not appearing when they should

apps/mobile/src/components/ContextSwitchMenu/ContextSwitchMenu.js
```diff
@@ -158,7 +161,7 @@ function ContextRow ({
             // (selected || bottomItem) && 'text-foreground'
           )}
         >
-          {context?.name}
+          {context?.name || 'No Name'}
```
tibetsprague, line 164: you think its better to show No Name? what if there is a flash of a moment where the name isnt loaded yet? id rather have it be empty than flash No Name

apps/mobile/src/screens/CreateGroup/CreateGroup.js
```diff
@@ -51,6 +51,16 @@ export default function CreateGroup ({ navigation }) {
     })
   }
 
+  // Cleanup effect - clear store if component unmounts while submitting
+  useEffect(() => {
+    return () => {
+      // If we're unmounting while submitting, clear the store immediately
+      if (isSubmitting) {
+        clearStore()
```
tibetsprague, line 59: what does clearStore do?

apps/mobile/src/screens/CreateGroup/CreateGroup.js
```diff
@@ -69,45 +79,48 @@ export default function CreateGroup ({ navigation }) {
...
-            />
-            )
-          : (
-            <ButtonNW
-              text={t('Lets Do This!')}
-              onPress={() => setSubmit(true)}
-              className='border-2 border-foreground/50'
-              textClassName='text-foreground'
-            />
+      {!isSubmitting && (
```
tibetsprague, line 82: why do we hide this stuff while submitting?

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
```diff
@@ -45,22 +48,63 @@ export const CreateGroupReview = React.forwardRef((_props, ref) => {
 
           if (data?.group) {
             trackWithConsent(AnalyticsEvents.GROUP_CREATED, {}, currentUser, !currentUser)
-            clearStore()
             changeToGroup(data.group.slug, { skipCanViewCheck: true })
+            clearStoreTimeoutRef.current = setTimeout(() => {
+              clearStore()
+            }, 5000)
```
tibetsprague, line 54: why a five second delay?

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
```diff
@@ -45,22 +48,63 @@ export const CreateGroupReview = React.forwardRef((_props, ref) => {
...
+          setIsSubmitting(false)
           setError(e.message)
         }
       })()
     }
   }, [submit])
 
+  useEffect(() => {
+    return () => {
+      if (clearStoreTimeoutRef.current) {
```
tibetsprague, line 69: why all this stuff?

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
```diff
@@ -45,22 +48,63 @@ export const CreateGroupReview = React.forwardRef((_props, ref) => {
...
+        </TouchableOpacity>
+      </View>
+    )
+  }
+
   return (
     <>
       <View className='mb-5'>
-        <Text className='text-foreground text-xl font-bold pb-2.5'>{t('Everything look good?')}</Text>
+        <Text className='text-foreground text-xl font-bold pb-2.5'>{t('Everything look good')}</Text>
```
tibetsprague, line 107: why remove ? seems like we want it

apps/mobile/src/screens/CreateGroup/CreateGroupUrl.js
```diff
@@ -12,11 +12,23 @@ export const BASE_STRING = 'hylo.com/groups/'
...
+    .replace(/^-|-$/g, '') // Remove leading/trailing hyphens
+    .substring(0, 40) // Limit to 40 characters
+}
+
 export default function CreateGroupUrl ({ navigation }) {
   const { t } = useTranslation()
   const { groupData, updateGroupData, setDisableContinue } = useCreateGroupStore()
   const [error, providedSetError] = useState()
-  const [groupSlug, setGroupSlug] = useState(groupData?.slug)
+  const [groupSlug, setGroupSlug] = useState(groupData?.slug || nameToSlug(groupData?.name || ''))
```
tibetsprague, line 31: what is this about?

packages/graphql/mutations/createGroupMutation.js
```diff
@@ -14,6 +14,16 @@ export default gql`
           person {
             id
           }
+          group {
```
tibetsprague, line 17: if this is the real fix then can we just do this?

packages/hooks/useCurrentUser.js
```diff
@@ -9,5 +10,18 @@ export default function useCurrentUser (useQueryArgs = {}) {
     return null
   }
 
-  return [{ currentUser: data?.me, fetching, error, stale }, queryCurrentUser]
+  // Memoized validation of memberships to ensure group property is always an object
+  const validatedCurrentUser = useMemo(() => {
```
tibetsprague, line 14: do we need this still after the fix above? just adds confusion

tibetsprague (changes requested):

apps/mobile/src/screens/CreateGroup/CreateGroup.js
thomasgwatson: Empties/resets the createGroup store.

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
thomasgwatson: basically a timeout on stalled creation mutations

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
thomasgwatson: Wanting to ensure that the clearStore function runs, even if the navigation action happens

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
tibetsprague: we should have a policy about this. we use ? in keys all the time right now and i think its confusing to have most keys be the full text but certain ones be almost the full text but not quite. im ok with keys that are single strings referencing long text. for right now i would recommend we keep using ? in particular

apps/mobile/src/screens/CreateGroup/CreateGroupUrl.js
tibetsprague: oh wow, we were not doing that in the app?? good catch

packages/hooks/useCurrentUser.js
tibetsprague: fair enough. i wouldnt really call this validation, more presentation. Also it would just be hiding other bugs again if the group was blank, but thats ok, better than a crash

apps/mobile/src/screens/CreateGroup/CreateGroupReview.js
thomasgwatson: I think the team should collectively use i18n-ally ( https://marketplace.cursorapi.com/items/?itemName=lokalise.i18n-ally and of course it works fine in VS code) and avoid the use of special characters in general. There might be some plugin config that I need to share, if it doesn't "just work" on install.

`.`  `_` and `:` have special meaning in the keys of the locale files and `{{ }}` has special meaning in the values. other special characters block some of the functionality of the IDE plugin.

we could certainly sync up our rules for how keys are composed.

I would guess that we defo want:
- human readable keys (and i18n-ally helps with this while looking at lines of code in the IDE)
- Pretty direct quotes for shorter strings, with some sort of vibe of 'this is too long'
- we currently don't namespace... but in some mythic future we might want to do that

and then I would ask that we don't use special characters so that we get the full functionality from i18n-ally

thomasgwatson:
@tibetsprague I did remove the timeout; instead just fully relying on the cleanup function for when the screen unmounts.

tibetsprague (approved):

PR #903, opened 2025-09-09
Regularly check whether sockets need to be reconnected and fo so

tibetsprague:
Hopefully will fix issues of browser tabs left alone for a while losing connection to sockets

### DIFF
--- a/apps/web/src/client/websockets.js
+++ b/apps/web/src/client/websockets.js
@@ -24,6 +24,71 @@ if (isClient) {
   io.sails.environment = environment
   io.sails.reconnection = true
   socket = io.socket
+
+  // Heartbeat + self-heal reconnect
+  const startSocketHeartbeat = (() => {
+    let started = false
+    return () => {
+      if (started) return
+      started = true
+
+      console.log('startSocketHeartbeat sockets')
+
+      const HEARTBEAT_PATH = '/noo/heartbeat'
+      const INTERVAL_MS = 30000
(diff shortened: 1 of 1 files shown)


apps/web/src/client/websockets.js
```diff
@@ -24,6 +24,71 @@ if (isClient) {
...
   socket = io.socket
+
+  // Heartbeat + self-heal reconnect
+  const startSocketHeartbeat = (() => {
+    let started = false
+    return () => {
+      if (started) return
+      started = true
+
+      console.log('startSocketHeartbeat sockets')
```
thomasgwatson, line 35: do you wanna leave these in for prod? or make dev only?

apps/web/src/client/websockets.js
```diff
@@ -24,6 +24,71 @@ if (isClient) {
...
+
+      const HEARTBEAT_PATH = '/noo/heartbeat'
+      const INTERVAL_MS = 30000
+      const TIMEOUT_MS = 10000
+      let consecutiveFailures = 0
+
+      const forceReconnect = () => {
+        try {
+          console.log('forceReconnect sockets')
```
thomasgwatson, line 44: ditto for all console logs, as the first question above

thomasgwatson (approved):
lgtm

apps/web/src/client/websockets.js
tibetsprague: im going to leave in for now, but remove eventually

PR #1022, opened 2025-10-15
Feature/chat comment drafts

brodeur:
Take a look at drafts in pots, chats, comments!


tibetsprague:
just tested this more and right now its not saving comment drafts correctly, and its not loading drafts from their saved version correctly always. also it seems to sometimes create duplicate drafts of a new post in the same group which shouldnt be possible

apps/backend/api/graphql/makeSchema.js
```diff
@@ -243,50 +246,7 @@ export default async function makeSchema ({ req }) {
       Mutation: makeMutations({ fetchOne }),
       Subscription: makeSubscriptions(),
 
-      // Custom Type resolvers
```
tibetsprague, line 246: this is unrelated, just moving stuff around, long story, but is slightly cleaner

PR #1027, opened 2025-10-17
Improved TrackEditor and TrackHome to match the new funding round fun…

brodeur:
…ctionality for publishing and unpublishing. Also cleaned up the style of the HyloEditor


tibetsprague:
wrapped into spaces and views

PR #1028, opened 2025-10-21
Paid content integration

thomasgwatson:
I want to get initial feedback on the migration.

The stripe connect guide allows you to select a few different options to model your overall use-case and then it generates a prompt for a coding assistant. I tweaked the prompt a little more and its generated a lot of code-changes and a lot of TODOs. But so far it looks like a decent start.

I want to be fairly measured about pulling in the changes piece by piece tho, and starting with the migration stuff


apps/backend/migrations/20251020160838_paid-content-stripe.js
```diff
@@ -0,0 +1,93 @@
...
+    })
+    // Create content_access table to track all content access grants (paid and free)
+    // This supports both Stripe purchases and admin-granted free access
+    .createTable('content_access', function (table) {
+      table.bigIncrements('id').primary()
+      table.bigInteger('user_id').unsigned().notNullable().references('id').inTable('users')
+      table.bigInteger('group_id').unsigned().notNullable().references('id').inTable('groups')
+      table.bigInteger('product_id').unsigned().references('id').inTable('stripe_products')
+      table.bigInteger('track_id').unsigned().references('id').inTable('tracks')
+      table.integer('role_id').unsigned().references('id').inTable('groups_roles')
```
tibetsprague: what is the role for here?

tibetsprague (commented):
looks good so far

apps/backend/api/controllers/StripeController.js
```diff
@@ -0,0 +1,367 @@
...
+   */
+  handleAccountUpdated: async function (event) {
+    try {
+      const account = event.data.object
+      if (process.env.NODE_ENV === 'development') {
+        console.log(`Account updated: ${account.id}`)
+      }
+
+      // Find the group with this Stripe account ID
+      const group = await Group.where({ stripe_account_id: account.id }).fetch()
```
tibetsprague: can multiple groups use the same stripe account? probably. i wonder if there is another identifier for like a sub-account or something like that? or maybe they just give separate account_ids for each one.

apps/backend/api/graphql/mutations/stripe.js
```diff
@@ -0,0 +1,400 @@
...
+      }
+
+      // Check if user is a steward/admin of the group
+      const membership = await GroupMembership.forPair(session.userId, groupId).fetch()
+      if (!membership || !membership.hasResponsibility(GroupMembership.RESP_ADMINISTRATION)) {
+        throw new GraphQLError('You must be a group administrator to create a connected account')
+      }
+
+      // Check if group already has a Stripe account
+      // TODO STRIPE: You may want to store the accountId in your Group model
```
tibetsprague: ok so it seems like each group will have its own "account"

apps/backend/api/graphql/schema.graphql
```diff
@@ -2482,6 +2492,8 @@ type TopicFollowSettings {
 
 type Track {
   id: ID
+  # Whether this track requires purchased access (paid content)
+  accessControlled: Boolean
```
tibetsprague: seems like we should call this the same thing we call it on group, so either both are paywall (or paywalled?) or both are accessControlled. i lean towards paywalled

docs/CONTENT_ACCESS_SUMMARY.md
```diff
@@ -0,0 +1,280 @@
...
+2. Trigger `content_access_expires_at_sync` fires automatically
+3. Function `sync_content_access_expires_at()` executes:
+   - **If track_id is NULL**: Updates `group_memberships.expires_at` (group-level or role-based purchase)
+   - **If track_id is set**: Updates `tracks_users.expires_at` (track-specific purchase, does NOT update group_memberships)
+   - **If role_id is set**: Updates `group_memberships_group_roles.expires_at` (role-specific purchase, also updates group_memberships if no track_id)
+
+**When access is revoked or expires:**
+1. Status changes to 'revoked' or 'expired' in `content_access`
+2. Trigger `content_access_expires_at_clear` fires
+3. Function `clear_content_access_expires_at()` executes:
```
tibetsprague: so it always has an expires_at value if content is accessible? and for lifetime we just put one far in the future?

docs/CONTENT_ACCESS_SUMMARY.md
```diff
@@ -0,0 +1,280 @@
...
+### Protection Against Expiration Overwriting
+
+**Scenario:** User has long-term group membership, buys short-term track access
+
+```javascript
+// Step 1: User buys 1-year group membership
+// Creates: content_access { user_id, group_id, expires_at: '2026-01-01', track_id: NULL }
+// Trigger updates: group_memberships.expires_at = '2026-01-01' ✓
+
+// Step 2: Same user buys 1-month access to Track A (3 months later)
```
tibetsprague: i think racks will always be lifetime?

tibetsprague (commented):
looking good so far!

docs/PAID_CONTENT_WORKFLOW.md
```diff
@@ -0,0 +1,325 @@
...
+     - "Please wait while we confirm your payment"
+     - "You will receive an email confirmation shortly"
+     - "Check your email for access details"
+
+2. **Email Confirmation**
+   - User receives confirmation email
+   - Email includes:
+     - Purchase confirmation details
+     - Links to access the content they purchased
+     - Instructions for accessing their new content
```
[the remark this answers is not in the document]
thomasgwatson: maybe "specific" is a better word than "special" here; a specific link to the content.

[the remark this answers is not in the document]
thomasgwatson: So, this is worth checking in about and there is still a bit of gray area based on exactly how the stripe hand-off goes:

Say, we are a user, in the stripe checkout and I hit purchase.
- I think then stripe redirects the user to the successUrl that we provide.
- Meanwhile, there is an async update between Stripe (via webhook) and us, letting us know that the payment worked out
- So, I am thinking that we made the successUrl pretty generic and simple and more or less just say: check your email

this saves us if there is a significant lag between transaction and webhook update.... but if it turns out that turn-around time is really fast (consistently?) then we are just adding another step (checking the email) for the user...

Any initial thoughts? Might just be something we figure out once we start seeing it implemented

apps/backend/api/controllers/StripeController.js
thomasgwatson: ooo right that is an interesting question. This is a webhook, so I will have to check if we can store a 'foreign key' of sorts from our db in their records for an account. I assume we can

PR #1058, opened 2025-10-28
Fix backend tests v3

thomasgwatson:
Ok, I cherry-picked good and/or straight-forward commits, added a few changes based on your review comments and basically bailed on all the code-swamps that I had previously tried to poke.

### DIFF
--- a/apps/backend/api/models/util/contextWidgets.js
+++ b/apps/backend/api/models/util/contextWidgets.js
@@ -96,7 +96,7 @@ export function replaceHomeWidget ({ widgets, newHomeWidgetId }) {
   // so by here the updatedWidgets array has removed the new home widget from its prior position and its prior peers are settled
 
   // if the old home widget is a chat, we need to move it to the top of the chats widget, otherwise we remove it from the menu
-  if (homeChild.type === 'chat' || homeChild.type === 'viewChat') {
+  if (homeChild.type === 'viewChat') {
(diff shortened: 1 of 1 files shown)


apps/backend/api/graphql/mutations/comment.js
```diff
@@ -11,9 +11,12 @@ export async function canDeleteComment (userId, comment) {
 
   const commentWithGroups = await comment.load('post.groups')
 
-  return commentWithGroups.relations.post.relations.groups.map(g =>
+  const hasResponsibilityPromises = commentWithGroups.relations.post.relations.groups.models.map(g =>
```
tibetsprague: still unclear on whether this was broken before, just making sure you are testing the functionality too

apps/backend/api/models/util/contextWidgets.js
```diff
@@ -96,7 +96,7 @@ export function replaceHomeWidget ({ widgets, newHomeWidgetId }) {
   // so by here the updatedWidgets array has removed the new home widget from its prior position and its prior peers are settled
 
   // if the old home widget is a chat, we need to move it to the top of the chats widget, otherwise we remove it from the menu
-  if (homeChild.type === 'viewChat') {
+  if (homeChild.type === 'chat' || homeChild.type === 'viewChat') {
```
tibetsprague: still think we can/should remove 'chat' completely, since it shouldnt be getting used anywhere anymore

tibetsprague (approved):

PR #1103, opened 2025-11-14
Improved search

tibetsprague:
Search defaults to searching only within current group but you can select search across all groups. Also clicking on a post in the search stream pops it open above the search results. 

Fixes https://github.com/Hylozoic/hylo/issues/902


thomasgwatson (approved):
lgtm but with one hanging TODO

PR #1122, opened 2025-12-04
Claude/chatrooms performance audit 011 c uq v nj6 hf v8htyn wm mjf s

brodeur:
Check out what Claude tried to do, may be good inspiration. Locally I am noticing a significant improvement over dev branch when I have the repos running side by side.


tibetsprague (commented):
testing this with the # of posts loaded at a time in the chat room set back to what it is on dev and this branch is not any faster than dev and in fact may be slightly slower. 😔

thomasgwatson:
Gonna close this one, since I think we have superseded it was a bunch of other optimizations

PR #1126, opened 2025-12-05
disable calendar view mode for phones

### LINKED ISSUE #1124: Calendar view is not responsive; does not work on small screens
When viewing on mobile web, the calendar view doesn't change its form to accommodate the smaller screen size. This is reproducible on any small screen platform.

This impacts mobile web and (likely impacts) tablet users. This view mode is the default for events, so it heavily impacts the Event view for each group.

As far as I understand, we are using a component library to render the fancy calendar. So how easy this is to fix will hinge on that library.

Possible solutions
- The library has a way to accommodate this and we need to configure it (yay)
- The library's primary components don't accommodate this but there are different components that do (conditional yay)
- If the library doesn't give us solutions:
- We hide calendar view altogether for smaller screens (sadness)
- We create an alternative rendering or list for events on smaler screens

Tagging @brodeur and @KevinTriplett for awareness but not with expectation that this needs to be prioritized per se

KevinTriplett:
Closes #1124 

Simple fix: disable calendar mode for small screens (isMobile.any) with minimal code change.

### DIFF
--- a/apps/web/src/components/StreamViewControls/StreamViewControls.js
+++ b/apps/web/src/components/StreamViewControls/StreamViewControls.js
@@ -1,4 +1,5 @@
 import { ArrowDownWideNarrow } from 'lucide-react'
+import isMobile from 'ismobilejs'
 import React, { useCallback, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useLocation } from 'react-router-dom'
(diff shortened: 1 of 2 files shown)


KevinTriplett:
I suggest we close the issue and add a new feature request for calendar view for mobile devices.

PR #1184, opened 2026-01-13
Scope mentions and topic selections

thomasgwatson:
Closes evo 1318

### DIFF
--- a/apps/backend/api/models/TagFollow.js
+++ b/apps/backend/api/models/TagFollow.js
@@ -210,7 +210,10 @@ module.exports = bookshelf.Model.extend(Object.assign({
         version: 'Redesign 2025',
         email: tagFollow.relations.user.get('email'),
         locale,
-        sender: { name: tagFollow.relations.group.get('name') + ' (via Hylo)' },
+        sender: {
+          name: tagFollow.relations.group.get('name') + ' (via Hylo)',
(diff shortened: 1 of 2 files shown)


apps/web/src/components/PostEditor/PostEditor.js
```diff
@@ -253,6 +253,28 @@ function PostEditor ({
...
+  // 1. Use selectedGroups if available (most common case - user has selected groups)
+  // 2. For public posts with no groups, allow all accessible topics/people (pass undefined)
+  // 3. Otherwise, fallback to currentGroup from route context
+  // Passing undefined allows backend to return all accessible topics/people based on user's group memberships
+  const groupIds = useMemo(() => {
+    if (selectedGroups && selectedGroups.length > 0) {
+      return selectedGroups.map(g => g.id).filter(Boolean)
+    }
+    // For public posts with no groups selected, allow all accessible topics/people
+    if (currentPost.isPublic && (!currentPost.groups || currentPost.groups.length === 0)) {
```
tibetsprague: but you always have to post to a group, so this should never really be used. I guess the question is what happens if you tag someone before adding a group, then add a group, or remove a group... I guess we still let them be tagged?

tibetsprague (approved):
woot!

PR #1199, opened 2026-01-19
Open draw on zoomed in cluster marker click, for  mobile map views

### LINKED ISSUE #1138: Commons - Map IOS
I cannot click on the map and drill down to the details.  

Mobile: Iphone 12 IOS Version 26.1 /On 

Image

thomasgwatson:
Closes #1138

### DIFF
--- a/apps/web/src/routes/MapExplorer/MapExplorer.js
+++ b/apps/web/src/routes/MapExplorer/MapExplorer.js
@@ -400,7 +400,10 @@ function MapExplorer (props) {
 
   const onMapClick = useCallback((info, e) => {
     if (info.objects) {
-      if (viewport.zoom >= 20 && hideDrawer) {
+      // On mobile, at high zoom levels (18-20), toggle drawer when clicking cluster
+      if (isMobileDevice() && viewport.zoom >= 18 && viewport.zoom <= 20) {
+        toggleDrawer()
+      } else if (viewport.zoom >= 20 && hideDrawer) {
         setHideDrawer(false)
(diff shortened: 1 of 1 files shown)


apps/web/src/routes/MapExplorer/MapExplorer.js
```diff
@@ -400,7 +400,10 @@ function MapExplorer (props) {
 
   const onMapClick = useCallback((info, e) => {
     if (info.objects) {
-      if (viewport.zoom >= 20 && hideDrawer) {
+      // On mobile, at high zoom levels (18-20), toggle drawer when clicking cluster
+      if (isMobileDevice() && viewport.zoom >= 18 && viewport.zoom <= 20) {
```
tibetsprague: im confused, we already do this at zoom >= 20 always, is the idea that in moblie you want that to happen sooner? also we should do the same thing we do below with setHideDrawer to false and then a trigger of resize, i think. or maybe we dont need the resize on mobile because its behind the drawer?

tibetsprague: @thomasgwatson thoughts?

thomasgwatson: The viewport is so small on mobile that its often the case that peeps can only see the map markers they are zooming in on. So its just extra zooms (often extra clicks on the markers), that feel redundant, when opening the drawer would show them the content they are actually trying to get to (the group or posts...)

tibetsprague (approved):

PR #1200, opened 2026-01-19
776 choose home view in group creation flow

### LINKED ISSUE #776: When creating a new group ask if they want #general or Stream to be home

thomasgwatson:
Closes #776 

Screenshot 2026-01-19 at 4 48 11 PM


apps/backend/api/models/Group.js
```diff
@@ -666,16 +669,43 @@ module.exports = bookshelf.Model.extend(merge({
...
-      group_id: this.id,
-      type: 'viewChat',
-      view_chat_id: generalTag.id,
-      parent_id: homeWidget.id,
-      order: 1,
-      created_at: new Date(),
-      updated_at: new Date()
-    }).save(null, { transacting: trx })
+    // Create home view widget based on homeView setting
+    // Only create general chat widget if homeView is CHAT (default)
```
tibetsprague: not sure about this. to discuss

apps/backend/api/models/Group.js
```diff
@@ -1032,13 +1086,22 @@ module.exports = bookshelf.Model.extend(merge({
     // XXX: for now groups by default cannot post to public on production
     attrs.allow_in_public = process.env.NODE_ENV === 'development'
 
+    // Extract homeView from data and add to settings (defaults to 'CHAT' if not provided)
+    const homeView = data.home_view || data.homeView || 'CHAT'
+    const defaultSettings = {
+      allow_group_invites: false,
+      agreements_last_updated_at: null,
+      public_member_directory: false,
+      homeView
```
tibetsprague: but do we update this when they change their home view? not sure we need to (or should) store it in settings, unless we do that. Though i guess it could be useful to see what they originally selected? To discuss

apps/backend/api/models/Group.js
thomasgwatson: Ahh this is inaccurate commentary, I changed it to ensure the general chat is ALWAYS created. Its just whether its the home view or unordered from the get-go

PR #1244, opened 2026-02-10
Make safe insets respect themeing; tighten bottom inset on iOS

### LINKED ISSUE #1242: Safe Area themeing for mobile
Front insert: emit theme background and style
bottom insert: conditionally ios => make shorter, also see if color from theme can be replicated

edge-to-edge

thomasgwatson:
Closes #1242 

This change ensures that themeing changes are tracked by mobile, and that the safe insets are styled with respect to those themes


apps/mobile/src/style/themes.js
```diff
@@ -0,0 +1,95 @@
...
+    const k = (n + h / 30) % 12
+    const color = l - a * Math.max(Math.min(k - 3, 9 - k, 1), -1)
+    return Math.round(255 * color).toString(16).padStart(2, '0')
+  }
+  return `#${f(0)}${f(8)}${f(4)}`
+}
+
+// Theme definitions — background and foreground HSL values per theme/variant.
+// Mirrors apps/web/src/themes/index.js
+const themeColors = {
```
tibetsprague: Would be ideal to not duplicate this code, but not sure it’s worth the effort to make it shared right now

PR #1323, opened 2026-04-08
Don't load custom views on initial group load

tibetsprague:
Will decrease initial load time of each group

Load only when viewing the custom view, or the group settings


thomasgwatson (approved):
Looks good to me

PR #1330, opened 2026-04-14
Extra skeletons

### LINKED ISSUE #1328: Dont show stream skeleton loader for Map, Groups, Group Settings, Calendar...
It's a bit confusing to show the stream skeleton when we are loading a page that is not stream or chat. The ones i know of now are:
- Map
- Groups
- Group Settings
- Calendar

but there may be more. 
Can we have different skeletons? 
Or just go back to loading spinner

thomasgwatson:
Closes #1328


tibetsprague (approved):
noice. one other place i noticed that would be good to do is when you go from the landing page at hylo.com to the signup/login page there is a brief skeleton that looks weird before it lands on the login form

PR #1332, opened 2026-04-16
1331 posts not preloading

### LINKED ISSUE #1331: Chat "new post" indicator non-responsive
Notifications are periodically updated in the app. And chatrooms will have their new-post counter increased. but the new posts are not available.... they are either not being passed via the sockets, or they are not otherwise polled and pre-loaded. But then also the UI for scrolling down is unresponsive and thus users have to manually refresh their content to see what the notifications are.

This is visible on both web and mobile (via the webview).

Image

thomasgwatson:
Closes #1331


apps/web/src/routes/ChatRoom/ChatRoom.js
```diff
@@ -352,7 +369,7 @@ export default function ChatRoom (props) {
     return () => {
       socket.off('newPost', handleNewPostReceived)
     }
-  }, [topicName])
+  }, [socket, handleNewPostReceived])
```
thomasgwatson, line 370: handleNewPostReceived now regenerates for changes in either topicName or Group.id

apps/web/src/routes/ChatRoom/ChatRoom.js
```diff
@@ -821,8 +843,17 @@ const StickyFooter = ({ context }) => {
         right: 50
       }}
     >
-      {(location.bottomOffset > 200 || context.newPostCount > 0) && (
-        <>
+      <div className='flex flex-col items-center gap-2'>
+        {showLoadingPulse && (
```
thomasgwatson, line 866: Just a little non-intrusive loading indicator

thomasgwatson:
I've deployed these changes to staging; lets see if we can test there

thomasgwatson:
These changes now include fixes to the "scroll to the latest", which seemed to be broken on prod

apps/web/src/routes/ChatRoom/ChatRoom.js
```diff
@@ -291,15 +291,41 @@ export default function ChatRoom (props) {
...
     // Incrementally fetch remaining future pages
     while (true) {
       const fetched = await fetchPostsFuture(offset, { first: INITIAL_POSTS_TO_LOAD }, true)
+      totalFetched += fetched || 0
       if (!fetched || fetched < INITIAL_POSTS_TO_LOAD) break
       offset += fetched
     }
-  }, [dispatch, location, topicFollow?.newPostCount, fetchPostsFuture, postsFuture?.length])
+
+    // Worst-case backstop: if we expected new posts but didn't fetch any, drop cached
```
tibetsprague: so we could remove this if we stop caching queryResults?

tibetsprague (commented):
looks ok, just want to test

apps/web/src/routes/ChatRoom/ChatRoom.js
tibetsprague: also this was happening because we were including actions and submissions in the new post count but they dont actually show in the chat, so then this would trigger a bunch of resets we dont actually want. so im not sure we should include this

tibetsprague: im actually worried this one will cause issues if the new post count is stuck because of a bug, it will keep doing an extra refresh

thomasgwatson: I just removed it

apps/web/src/routes/ChatRoom/ChatRoom.js
```diff
@@ -821,15 +857,27 @@ const StickyFooter = ({ context }) => {
...
+              <div className='h-1.5 w-1.5 rounded-full bg-foreground/35' />
+              <div className='h-1.5 w-1.5 rounded-full bg-foreground/25' />
+            </div>
+          </div>
+        )}
+        {showJumpButton && (
           <button
             className='relative flex items-center justify-center bg-background border-2 border-foreground/15 rounded-full w-8 h-8 text-foreground/50 hover:text-foreground'
             onClick={() => {
+              context.messageListRef?.current?.scrollToItem({ index: 'LAST', align: 'end', behavior: 'auto' })
```
tibetsprague: then it does a two part scroll though right? why do you think thats better?

tibetsprague (approved):
i guess this is fine, though scroll to latest is working for me on prod. what behavior were you seeing?

apps/web/src/routes/ChatRoom/ChatRoom.js
thomasgwatson: Right, if the other changes were actually what was causing the weird UX, lets just reset the scrolling changes here and see how we go.

PR #1337, opened 2026-04-17
Remove query results from redux-persist

tibetsprague:
Also fix flicker of no posts in chat room when loading
Also revert a recent change that broke a fix that resets a stuck new post count in a chat room.


thomasgwatson (approved):
lgtm

PR #1367, opened 2026-05-05
Avoid public comments leading to badly formed urls in notifications

### LINKED ISSUE #1357: Notification routing bug
A user describes:

"I would comment on a post in the public stream. Then when I would get a response to one of my comments and clicked the notification, it took me to the index location of the original post, which was in a members-only group.  I don't know how a closed group could have posts that are visible from the public stream, but that's what seems to have happened. "

So we are sending notifications to users, including a group association, when they don't have that group membership

Likely a backend fix

thomasgwatson:
Closes #1357


tibetsprague (approved):
Nice one

PR #1371, opened 2026-05-07
Refine what hylo links opne the android app

### LINKED ISSUE #1336: Android: Hylo app intercepts OAuth /oauth/login/* links, breaks third-party OIDC sign-in on mobile
Companion to #1335 (iOS AASA fix). The Android side of the same problem needs a mobile-team decision about scope.

## Problem

Third-party OIDC clients using Hylo as an auth provider (e.g. [castalia.one](https://castalia.one), which consumes `/noo/oauth/auth`) can't complete sign-in on **Android** when the Hylo native app is installed. The flow:

1. Consumer app redirects the user to `https://www.hylo.com/noo/oauth/auth?...` — correctly excluded from Hylo's OIDC protocol endpoints.
2. Hylo server 303s to `/noo/oidc/interaction/{id}` — also excluded.
3. That page redirects the browser to `/oauth/login/{id}?name=<ClientName>` — the user-facing login UI.
4. **Android App Links** intercept `/oauth/login/*` because Hylo's `assetlinks.json` uses:
   ```json
   "relation": ["delegate_permission/common.handle_all_urls"]
   ```
   …which claims **every** URL on `www.hylo.com` and `staging.hylo.com` with no path filter. The current intent filter in `apps/mobile/android/app/src/main/AndroidManifest.xml`:
   ```xml
   <intent-filter android:autoVerify="true">
     <action android:name="android.intent.action.VIEW" />
     <category android:name="android.intent.category.DEFAULT" />
     <category android:name="android.intent.category.BROWSABLE" />
     <data android:scheme="https" android:host="www.hylo.com" />
     <data android:scheme="https" android:host="staging.hylo.com" />
   </intent-filter>
   ```
   …has no `android:pathPrefix` / `android:pathPattern`, so it matches the full host.

5. The Hylo app launches, has no handler for the OAuth login route, and the user is stranded.

## Why iOS was fixable but Android is harder

iOS AASA supports `"exclude": true` at the path-component level — PR #1335 simply appends a `/oauth/*` exclusion before the `"/": "*"` catch-all and we're done.

Android App Links don't have an exclusion primitive. The Android fix has to **enumerate the paths the Hylo app actually wants to handle** and replace the broad intent-filter with one (or several) scoped by `android:pathPrefix` / `android:pathPattern`. For example, if the Hylo Android app deep-links to group pages, member profiles, etc., the fix would look something like:

```xml
<intent-filter android:autoVerify="true">
  <action android:name="android.intent.action.VIEW" />
  <category android:name="android.intent.category.DEFAULT" />
  <category android:name="android.intent.category.BROWSABLE" />
  <data android:scheme="https" android:host="www.hylo.com" />
  <data android:scheme="https" android:host="staging.hylo.com" />
  <data android:pathPrefix="/noo/login/token" />
  <data android:pathPrefix="/noo/login/jwt" />
  <data android:pathPrefix="/groups/" />
  <data android:pathPrefix="/m/" />
  <!-- …other first-party deep-link paths here, omitting /oauth/* and /noo/oauth/* -->
</intent-filter>
```

Picking that list correctly needs mobile-team knowledge about what URLs the Android app is expected to handle natively, so I didn't include this change in #1335.

## Requested action

A mobile-team maintainer to:
1. Enumerate the set of `www.hylo.com` paths the Android app should claim.
2. Update `apps/mobile/android/app/src/main/AndroidManifest.xml`'s HTTPS intent-filter with `android:pathPrefix` entries that omit `/oauth/*` and `/noo/oauth/*`.
3. (Optionally) also update `apps/web/public/.well-known/assetlinks.json` if the Digital Asset Links protocol ever adds path-scoped delegation; today it only supports host-level `handle_all_urls`, so the enforcement really lives in the manifest.

Happy to open the Android PR once the path list is confirmed.

## Background

Same debugging session as #1335. iOS PR covers the case where the Hylo iOS app intercepts consumer OAuth. This is the Android twin of that issue.

thomasgwatson:
Closes #1336

### DIFF
--- a/apps/mobile/android/app/src/main/AndroidManifest.xml
+++ b/apps/mobile/android/app/src/main/AndroidManifest.xml
@@ -57,13 +57,92 @@
         <action android:name="android.intent.action.MAIN" />
         <category android:name="android.intent.category.LAUNCHER" />
       </intent-filter>
-      <!-- Intent filter for HTTPS URLs -->
+      <!--
+        Scoped App Links: each <data> line is OR'd. Omit /oauth/* and /noo/oauth/* so third-party
+        OIDC and Hylo-as-provider flows stay in the browser. Omit /.well-known/* (OIDC discovery).
+        Omit /noo/oidc/* (interaction API — browser). Only /noo/login* is listed under /noo (magic
+        links + OAuth GET hops); other /noo/* routes are API/session plumbing, not link targets.
+        When adding first-party URL families, add path rules for both hosts below.
+        Short first segments use android:path (exact) plus pathPrefix "/segment/" to avoid stealing
+        unrelated paths (e.g. /post vs /postal). Omit marketing/static paths (same idea as
+        navigation/linking staticPages — /about, /help, /terms, …) so those URLs stay in the browser.
+      -->
       <intent-filter android:autoVerify="true">
         <action android:name="android.intent.action.VIEW" />
         <category android:name="android.intent.category.DEFAULT" />
         <category android:name="android.intent.category.BROWSABLE" />
-        <data android:scheme="https" android:host="www.hylo.com" />
-        <data android:scheme="https" android:host="staging.hylo.com" />
+        <data android:scheme="https" android:host="www.hylo.com" android:path="/" />
+        <data android:scheme="https" android:host="staging.hylo.com" android:path="/" />
+        <data android:scheme="https" android:host="www.hylo.com" android:path="/all" />
+        <data android:scheme="https" android:host="staging.hylo.com" android:path="/all" />
+        <data android:scheme="https" android:host="www.hylo.com" android:pathPrefix="/all/" />
+        <data android:scheme="https" android:host="staging.hylo.com" android:pathPrefix="/all/" />
+        <data android:scheme="https" android:host="www.hylo.com" android:path="/create-group" />
+        <data android:scheme="https" android:host="staging.hylo.com" android:path="/create-group" />
+        <data android:scheme="https" android:host="www.hylo.com" android:pathPrefix="/create-group/" />
+        <data android:scheme="https" android:host="staging.hylo.com" android:pathPrefix="/create-group/" />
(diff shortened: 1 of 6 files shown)


tibetsprague (approved):
seems like this needs some thorough testing, but nice

thomasgwatson:
Yeah I'm not going to merge straight away, want to build an android staging version and play around

thomasgwatson:
Doesn't seem to work for the OIDC links. They open in the app still

thomasgwatson:
So this PR does 1 and now 3

1. Ship the scoped App Links fix and verify the merged APK (primary)
Ensure the release APK’s merged AndroidManifest has no host-only <data android:scheme="https" android:host="www.hylo.com" /> (same for staging).
Keep /oauth/* and /noo/oauth/* off the allowlist (as you designed).
That’s the main fix: Hylo is not a candidate handler for those URLs, so Chrome keeps the flow.

2. Server / product: OIDC on a URL the app never claims (belt and suspenders)
If you want zero dependence on manifest merge quirks:

Put the browser OIDC authorize / login UI on another origin that the app does not register, e.g. https://auth.hylo.com/... or a path on a host the Hylo app doesn’t link to.
Consumer apps would use that authorize URL; the Hylo app would only claim www.hylo.com (or whatever you keep for in-app deep links).
This is the most robust long-term pattern many products use (“auth on accounts.* / login.*”).

3. App fallback: if Hylo opens an OAuth URL, bounce to the browser (safety net)
If a bad merge or edge case still opens Hylo for /oauth/..., MainActivity / linking can detect that path and call Linking.openURL(sameUrl) (or an intent chooser) so the user lands in Chrome. That doesn’t fix verification or marketing links, but it unblocks OIDC when the app stole the intent.

4. User-facing workaround (last resort)
Settings → Apps → Hylo → Open by default → clear “Open supported links” / pick “Ask every time” / open link in browser. Document this only if you must.

tibetsprague (approved):

PR #1372, opened 2026-05-07
Post link improvements

tibetsprague:
Change how we wrap text in posts so we can display full links

Make sure links like google.com (without https://) used in posts open correctly
They were opening up as internal Hylo posts

Add more sophisticated Link editing to post editor
Remove link becomes Edit Link. Can edit text and link URL


thomasgwatson (approved):
lgtm

PR #1374, opened 2026-05-07
Fix roll badge not saving

### LINKED ISSUE #1373: [Roles] Changing the emoji on a role is not saving
from WTR: "It looks like it should be possible to change the emojis used for our role badges but it's not allowing me to switch them out. It pulls up the popup but when I click on the emoji nothing happens. "

I tested and was able to select a new emoji for a role, but when I went back into the group to look at it, the old emoji was displayed.

https://hylo.canny.io/admin/feedback/bugs-feature-requests/p/roles-changing-the-emoji-on-a-role-is-not-saving

thomasgwatson:
Closes #1373

### DIFF
--- a/apps/web/src/routes/GroupSettings/RolesSettingsTab/RolesSettingsTab.js
+++ b/apps/web/src/routes/GroupSettings/RolesSettingsTab/RolesSettingsTab.js
@@ -22,6 +22,7 @@ import {
   fetchStewardSuggestions,
   clearStewardSuggestions
 } from './RolesSettingsTab.store'
+import { fetchGroupSettings } from '../GroupSettings.store'
 import getPerson from 'store/selectors/getPerson'
 import SettingsSection from '../SettingsSection'
 import EmojiPicker from 'components/EmojiPicker'
@@ -50,7 +51,7 @@ const validateRole = ({ name, emoji }) => {
   return true
 }
 
-function RolesSettingsTab ({ group, commonRoles }) {
+function RolesSettingsTab ({ group, commonRoles, slug }) {
   const dispatch = useDispatch()
   const suggestions = useSelector(state => state.RoleSettings.map(personId => getPerson(state, { personId })))
   const { t } = useTranslation()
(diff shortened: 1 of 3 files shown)


apps/web/src/routes/GroupSettings/RolesSettingsTab/RolesSettingsTab.js
```diff
@@ -225,13 +226,21 @@ function RoleRow ({
   const isDraftRole = active === ''
   const inactiveStyle = (!active && !isDraftRole && !isCommonRole) ? styles.inactive : ''
   return (
-    <div className={cn('bg-foreground/5 rounded-lg my-4 pb-2', inactiveStyle)}>
```
thomasgwatson, line 228: Fixed a styling bug while I was here

apps/web/src/routes/GroupSettings/RolesSettingsTab/RolesSettingsTab.js
```diff
@@ -114,9 +115,9 @@ function RolesSettingsTab ({ group, commonRoles }) {
     const role = { ...roles[i] }
     if (validateRole(role)) {
       dispatch(addGroupRole({ ...role, groupId: group?.id })).then((response) => {
-        const updatedRoles = roles
-        updatedRoles[i] = { ...response.payload.data.addGroupRole }
-        setRoles(updatedRoles)
+        const saved = response.payload.data.addGroupRole
+        setRoles(prev => prev.map((r, idx) => idx === i ? { ...saved } : r))
+        if (slug || group?.slug) dispatch(fetchGroupSettings(slug || group.slug))
```
tibetsprague, line 120: why do we have to fetch the settings again?

tibetsprague (approved):
seems fine, though I dont understand it

apps/web/src/routes/GroupSettings/RolesSettingsTab/RolesSettingsTab.js
thomasgwatson: To make sure the redux store is accurate, not just the local state.

tibetsprague: ideally that would be handled by a reducer, with an optimistic update, but also, this is ok since this is not happening a lot

PR #1376, opened 2026-05-08
separate auth state from currentUser state

thomasgwatson:
Obvs needs testing on staging as well. Might be simpler to have a separate staging-auth branch for testing this, since I don't necessarily want to cross-pollinate it with all the other staging changes, for testing simplicity sake. Lets line up a dedicated time to test for next week


tibetsprague:
wait remind what this is for?

thomasgwatson:
> wait remind what this is for?

1) Cleaner separation of auth from user state
2) foundational work for any persistent/caching progress

thomasgwatson:
Logging a few of the design decisions here:

Where it's still entangled (important, by current design):

getAuthState / getAuthorized / getSignupInProgress / getSignupComplete still combine getAuthSessionAuthenticated AND getMe, because the signup-state machine inherently reads user fields (emailValidated, hasRegistered, settings.signupInProgress). So "authorized" still requires the me ORM row to be present.

Today this is consistent because CHECK_LOGIN populates the ORM me and flips authSession to Authenticated in the same action — they can't diverge at runtime.
**This is exactly the seam that will bite redux-persist.** On rehydrate you could have authSession = Authenticated with the me row not yet present → getAuthorized returns false → user briefly bounced toward public/login routes (the "disaster" from last time). When you do the persist work, getAuthorized should depend only on authSession (or you persist a minimal me), and the signup gating should be split out from the authorize gate.


Can we decouple getAuthState/getAuthorized from getMe? Yes
The authorization decision only depends on two small, stable auth facts — emailValidated and hasRegistered — not on the rich profile data in the ORM me. Those two flags already come back on every auth action (login, verifyEmail, register, checkLogin). So the clean move is to store them in the authSession slice (the thing redux-persist will persist), and derive getAuthState/getAuthorized from authSession alone. getSignupInProgress/getSignupComplete keep reading settings.signupInProgress from me, exactly as you want.

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
```diff
@@ -252,7 +252,7 @@ export default function ContextMenu (props) {
                 : isMyContext
                   ? (
                     <div className='MyHomeHeader relative flex flex-col justify-end p-2 bg-cover h-[190px] shadow-md'>
-                      <div className='absolute inset-0 z-10 bg-cover bg-center' style={{ ...bgImageStyle(currentUser.bannerUrl || '/default-user-banner.svg'), opacity: 0.8 }} />
+                      <div className='absolute inset-0 z-10 bg-cover bg-center' style={{ ...bgImageStyle(currentUser?.bannerUrl || '/default-user-banner.svg'), opacity: 0.8 }} />
```
tibetsprague, line 255: all these parts that allow no user still creep me out. shouldnt we only load AuthLayoutRouter once we have a user?

apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -674,7 +678,7 @@ export default function AuthLayoutRouter (props) {
 
   /* First time viewing a group redirect to welcome page if it exists, otherwise home view */
   // XXX: this is a hack, figure out better way to do this
-  if (currentGroupMembership && !get('lastViewedAt', currentGroupMembership)) {
+  if (currentUser && currentGroupMembership && !get('lastViewedAt', currentGroupMembership)) {
```
tibetsprague, line 698: do we really need all these allowances?

apps/web/src/store/reducers/authSession.js
```diff
@@ -0,0 +1,137 @@
...
+    if (!action?.payload?.data) return state
+
+    if (!me) {
+      return anonymousSession()
+    }
+
+    return authenticatedSession(me)
+  }
+
+  if (action.type === LOGOUT || action.type === CLEAR_AUTH_SESSION) {
```
tibetsprague: why do we need logout and clear_auth_session both?

apps/web/src/store/reducers/resetStore.js
```diff
@@ -10,7 +11,14 @@ export const KEYS_PRESERVED_ON_RESET = [
 
 export default function (state = null, action) {
   if (action.type === LOGOUT && !action.error) {
```
tibetsprague, line 12: so we handle logout in two different reducers? why?

apps/web/src/store/selectors/getSignupState.js
```diff
@@ -1,5 +1,10 @@
 import { createSelector } from 'reselect'
-import getMe from 'store/selectors/getMe'
+import {
```
tibetsprague, line 2: why do we need both getAuthSession and getAuthState? that is confusing to me

apps/web/src/store/selectors/getAuthState.js
tibetsprague, line 11: i dont actually think any of this has to do with authorization, authorization is about group membership, roles, admins, etc. right? this is all about authentication i think

apps/web/src/store/constants.js
```diff
@@ -197,7 +197,9 @@ export const SORT_NAME = 'name'
 export const SORT_NEAREST = 'nearest'
 export const SORT_SIZE = 'size'
 export const SET_CONFIRM_BEFORE_CLOSE = 'SET_CONFIRM_BEFORE_CLOSE'
+export const SET_AUTH_SESSION = 'SET_AUTH_SESSION'
 export const SET_STATE = 'SET_STATE'
+export const CLEAR_AUTH_SESSION = 'CLEAR_AUTH_SESSION'
```
tibetsprague: i dont actually see this used anywhere

apps/web/src/store/constants.js
```diff
@@ -197,7 +197,9 @@ export const SORT_NAME = 'name'
 export const SORT_NEAREST = 'nearest'
 export const SORT_SIZE = 'size'
 export const SET_CONFIRM_BEFORE_CLOSE = 'SET_CONFIRM_BEFORE_CLOSE'
+export const SET_AUTH_SESSION = 'SET_AUTH_SESSION'
```
tibetsprague: i dont see this used anywhere

tibetsprague (commented):
i think i need to be walked through with a lot of detail to understand why we need this. the idea is when we bring back redux persist we can persist the me stuff but not the auth stuff so we do a quick auth check first, to see if you are logged in, and if you are then we can reuse the existing me slice (while we reload it from the database)? i think that makes sense to me... 🤔

apps/web/src/store/reducers/authSession.js
thomasgwatson: I mentioned to cursor that this was prep work for redux persist and it did this extra prep. So `clear_auth_session` is not currently in use but it would be a mechanism to reset the auth but not dump the rest of the redux store.

apps/web/src/store/constants.js
thomasgwatson: Yep, address in the comment above. This was pre-emptive prep for redux persist that I didn't ask to be done but it happpppened.

We can re-add them for when we actually do persist

apps/web/src/store/selectors/getAuthState.js
thomasgwatson: Ehh, this is a four year old comment from Loren.

One can be authenticated as a user but not really authorised to do anything if you haven't verified your email, so I think that was his distinction

apps/web/src/store/selectors/getSignupState.js
thomasgwatson: Ok, this is bad naming.

This PR introduces something, `getAuthSession`, that shows up poor naming of something else

`getAuthSession` (the slice) vs `getAuthState` (the derived funnel) sound interchangeable, but getAuthState's values are literally signup stages — the file header even calls it "Signup State." If you want to make the distinction obvious, we could rename getAuthState/AuthState to something like getSignupState/SignupState (or getAuthFunnelState).

I'll do a re-name

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/ContextMenu.jsx
thomasgwatson: A lot of these are just straight-forward null guards. I asked to ensure uses of currentUser usage was more robust.

thomasgwatson:
I've added the changes from your `improvements-to-mobile-auth` branch to this one as well, so they are integrated

thomasgwatson:
@tibetsprague ok, I want to release this.

It needs to go out on Web first, then I can cut the mobile releases.

So I want to merge, and I'm fine doing a web release and the ceremony there but want to double check whether you want to go over this PR again

apps/web/src/routes/AuthLayoutRouter/AuthLayoutRouter.js
tibetsprague: if we dont have currentUser here we should probably redirect somewhere else?

tibetsprague (approved):
ok lets do it

PR #1379, opened 2026-05-12
Better handling of isolated post detail views in mobile web

### LINKED ISSUE #1377: When viewing an individual post URL on mobile the menu navigation is wack
Clicking the back button opens the global nav but no context menu which looks weird. Should we put them in a context? Also then clicking back on the post does not close the nav menu. So its very hard to know what to do

- [ ]Let's do a better job routing to the right context. Default to my context?
- [ ] Also My should default to all My Groups Content instead of My Posts

thomasgwatson:
Closes #1377

### DIFF
--- a/apps/backend/scripts/seed-e2e-baseline.js
+++ b/apps/backend/scripts/seed-e2e-baseline.js
@@ -203,6 +203,107 @@ async function main () {
       [publicPostId, publicGroupId]
     )
 
+    /** Groups + posts for Playwright post-detail close navigation (see apps/web/e2e/authenticated.post-detail.spec.js). */
+    const outsiderGroupRes = await client.query(
+      `INSERT INTO groups (
+        active, created_at, updated_at, name, slug, description,
+        visibility, accessibility, created_by_id, settings, num_members, allow_in_public
+      ) VALUES (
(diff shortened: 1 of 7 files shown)


tibetsprague:
im unclear how these changes fix the issues mentioned. All this seems to do is route somewhere different when you close the post. The issue I am seeing is when you press the "back" button in the upper toolbar on mobile, which doesnt close the post, it just opens up the global nav menu. Which then cant be closed except by figuring out how to swipe it back to the left. The main fix we talked about was having any notification URLs that may take you to /post/id instead put you in the /my context so there is always a context menu when you press the back button. Another fix would be if you somehow end up in the context-less place then pressing anywhere not on the global nav menu should close it and go back to the post. What does your fix do?

thomasgwatson:
Any exiting from `/posts/:postId` (when there is no context) now has a specific redirection based on whether the user has a membership, the post being public, etc.

It is meant to avoid the user ending up in the broken state. I want to actually test it on a device in staging, since there are more ways to mess up with the UI on an actual device than the stimulator

tibetsprague:
but this does not fix the back button thing still, since that isn't closing the post, its just opening the nav menu

thomasgwatson:
Right, sorted that out now, sorry

tibetsprague (approved):

Issue #731, opened 2025-07-30
Backend: Link/path is shared from backend with group I'm not a member of, even though I am permitted to see the post/content


thomasgwatson:
It looks like web might be handling this mis-direction from the backend by itself. See the web notif; it references a group I'm a part of:
Web screen shot of notif:
![Image](https://github.com/user-attachments/assets/2aaa4bfc-1975-4361-b6e6-8dd611de655c)

but due to the constraints of mobile nav; we cannot easily handle this mis-direction encoded into the path by the backend.

So my preference would be for the backend to avoid this misdirection; to check group membership and build the notification path base with a group that the user is actually a member of.

tibetsprague:
is this fixed by your recent change?

hylobot:
Kinda... Specifically it was for comment notifications still returning the
first group from the post in the url, even if the user was not a member.

So this *could* still be an issue?

tibetsprague:
lets see if this is still an issue

thomasgwatson:
pre-mobile-web cross-over, so it won't be replicated (the section of mobile nav that was complicating this is gone)

Issue #700, opened 2025-07-11
Clicking on a link to Reply in Group from a post email took me to "wrong" chat room?


tibetsprague:
https://www.hylo.com/groups/pha/chat/climatechange?postId=90673&?ctt=post_email&cti=7650&ctcn=Planetary+Health+Alliance+%28MAIN%29

Took me to climatechage chat room even though the post was posted in the general chat too. Which one should it have taken me too? The bigger issue is that the #climatechange chat room showed up as empty even though on refresh it had hell posts!

tibetsprague:
maybe fixed?

thomasgwatson:
Yeah, I suspect this one will have been ironed out by fixes to notification links and handling

PR #1388, opened 2026-05-20
1380 notification urls whyyyy

### LINKED ISSUE #1380: Push notification link not going to correct place report
More about the mobile issue WTR mentioned, I was just able to validate this:
- I get a push notification about a response to my post
- I click the notification, it opens the ios app and takes me to the wrong group (dead end)
- I go to the in-app notifications and click the notifcation, it takes me to the correct group and post and scrolls down to the new comment which is highlighted in yellow (yay)
- then I scroll back up - the UI goes black and reloads, taking me again back down to the yellow highlighted comment. it seems like this is just happening once - in the past it happening repeatedly so there has been some improvement!

thomasgwatson:
Closes #1380

Ok, so this requires a MOBILE deployment first, then backend, or we'll get into trouble.

I can do a sanity check on staging for basic functionality but won't really be able to check the full flow before pushing to prod :\

These are the associated bugs that could have been in play:

### Bug 1 — commentId is silently dropped from every push URL (Backend)
apps/backend/api/models/Notification.js line 268, and every other push sender that calls .pathname.

Notification.js
Lines 268-270
    const path = new URL(Frontend.Route.comment({ comment, group, post })).pathname
    const alertText = PushNotification.textForComment(comment, version, locale)
Frontend.Route.comment() delegates to Frontend.Route.post() with extraParams = 'commentId=XXX'. That produces a URL like:

https://www.hylo.com/groups/wtf/post/789?commentId=500
.pathname returns only /groups/wtf/post/789. The ?commentId=500 is in .search and is completely thrown away. The push notification URL never carries the commentId.

The fix is two characters:

const postURL = new URL(Frontend.Route.comment({ comment, group, post }))
const path = postURL.pathname + postURL.search
Every other push sender that does new URL(...).pathname should be audited for the same pattern — post push, event invitation push, etc.

### Bug 2 — hyloapp:/ + /path silently produces a malformed URL (Backend + Mobile)
apps/backend/api/services/OneSignal.js line 34

OneSignal.js
Lines 33-35
  // if (path) notification.app_url = 'hyloapp://groups/heart-orchard/stream/post/78041'
  if (path) notification.app_url = 'hyloapp:/' + path
path always starts with / (it comes from .pathname). So concatenation produces:

'hyloapp:/' + '/groups/wtf/post/789'  =  'hyloapp://groups/wtf/post/789'
The double slash tells every WHATWG-compliant URL parser (including react-native-url-polyfill) that what follows is an authority (host). So groups becomes the hostname and the actual path becomes /wtf/post/789.

This breaks the useOpenURL path:

// apps/mobile/src/hooks/useOpenURL.js line 35-41
const linkingURL = new URL(providedPathOrURL, DEFAULT_APP_HOST)
// For 'hyloapp://groups/wtf/post/789':
//   linkingURL.host     = 'groups'
//   linkingURL.pathname = '/wtf/post/789'   ← /groups/ prefix is gone
//   linkingURL.origin   = 'null'            ← custom scheme, never matches prefixes
Because origin is 'null', useOpenURL falls through to Linking.openURL(rawURL), which fires a Linking event. React Navigation's listener does correctly strip hyloapp:// → groups/wtf/post/789 and then getStateFromPath re-resolves that against DEFAULT_APP_HOST — so the catch-all does eventually produce the right WebView path. But it requires this extra Linking bounce, adds latency, and is completely invisible to any debugging.

More importantly, if useOpenURL is ever called directly (e.g., from an in-app deep link handler) without the Linking bounce, the truncated path /wtf/post/789 would produce a 404 WebView load — which is exactly the "wrong group, dead end" symptom.

Two-line backend fix — use a proper scheme with a dummy host, which produces a clean path on both parsers:

// OneSignal.js line 34
if (path) notification.app_url = 'hyloapp://hylo.com' + path
Then in mobile prefixes, the 'hyloapp://' entry also needs to be updated to 'hyloapp://hylo.com' so useOpenURL's origin check matches:

// apps/mobile/src/navigation/linking/index.js
export const prefixes = [
  DEFAULT_APP_HOST,
  'https://staging.hylo.com',
  'hyloapp://hylo.com',   // was 'hyloapp://'
]
With this, new URL('hyloapp://hylo.com/groups/wtf/post/789').origin = 'hyloapp://hylo.com', which is in prefixes, and pathname = /groups/wtf/post/789 ✓. useOpenURL can navigate directly without the Linking bounce.

### Bug 3 — Wrong group chosen for push URL (Backend)
apps/backend/api/models/Notification.js

For comment notifications, sendCommentPush uses post.groupForFrontendRouteForUser(reader.id):

Notification.js
Lines 262-268
  sendCommentPush: async function (version) {
    const comment = this.comment()
    const post = comment.relations.post
    const reader = this.reader()
    const group = await post.groupForFrontendRouteForUser(reader.id)
And groupForFrontendRouteForUser returns the first group in the post's group collection that the reader is a member of:

Post.js
Lines 181-195
  async groupForFrontendRouteForUser (viewerUserId) {
    ...
    const memberGroup = groups.models.find(g => userGroupIds.includes(g.id))
    return memberGroup || null
  },
For post notifications, sendPostPush uses Activity.groupIds()[0] — the first group in the post's groups relation:

Notification.js
Lines 248-259
  sendPostPush: async function (version) {
    const post = this.post()
    const groupIds = Activity.groupIds(this.relations.activity)
    ...
    // TODO: include all groups in the notification?
    return Group.find(groupIds[0])
Neither uses activity.group_id, which is the actual group context where the notification was generated (set from comment.group() → post.groups.models[0]). For a multi-group post, or if the DB collection order doesn't match the notification context, these can return a different group than the one the user was in when they saw the comment.

The correct source of truth is activity.relations.group (the group_id on the activity row). For sendCommentPush:

sendCommentPush: async function (version) {
  const comment = this.comment()
  const post = comment.relations.post
  const reader = this.reader()
  // Use the activity's group (the context where the notification was created)
  // and fall back to groupForFrontendRouteForUser if not available
  const activityGroup = this.relations.activity?.relations?.group
  const group = activityGroup || await post.groupForFrontendRouteForUser(reader.id)
  ...
}
### Bug 4 — No OneSignal notification-opened handler (Mobile)
apps/mobile/index.js

There is a foregroundWillDisplay handler, but no click (notification opened) handler:

index.js
Lines 147-158
      const foregroundWillDisplayHandler = event => {
        event.preventDefault()   // ← suppresses foreground display entirely
        console.log('foreground notification received:', event.notification)
      }
      OneSignal.Notifications.addEventListener('foregroundWillDisplay', foregroundWillDisplayHandler)
Without a click handler, OneSignal falls back to OS-level URL opening for tapped notifications. This means:

No logging when a notification is tapped
No Sentry tracking if navigation fails
The OS handles app_url before the JS layer is ready (cold start race)
Foreground notifications are silently swallowed (event.preventDefault without re-showing)
Add an opened handler:

const notificationOpenedHandler = event => {
  const { notification } = event
  const url = notification.launchURL || notification.additionalData?.path
  if (url) {
    openURL(url).catch(err => Sentry.captureException(err, { extra: { url } }))
  }
}
OneSignal.Notifications.addEventListener('click', notificationOpenedHandler)
This gives you explicit control, logging, and proper useOpenURL routing for the notification tap — bypassing the OS Linking bounce entirely.

### Bug 5 — Scroll-up causing black flash / reload
This one is a symptom of the WebView-over-navigation architecture. The sequence:

Push tap → WebView loads /groups/wtf/post/789 (no commentId due to Bug 1)
User sees the post without any highlighted comment, so they can't find what triggered the notification

User opens in-app notification → openURL('/groups/wtf/post/789?commentId=500') → originalLinkingPath changes → React Navigation dispatches a new navigate action to AuthRoot/Main

PrimaryWebView re-renders with the new originalLinkingPath and reloads the WebView URI — the web app hard-navigates to the post with commentId, Comment.jsx scrolls to the comment

User scrolls up → they're now reading the post from the top — but originalLinkingPath is still /groups/wtf/post/789?commentId=500 in navigation state. If any navigation state change triggers a re-render of PrimaryWebView (e.g., a Redux update, a URQL response, AuthRootNavigator's user fetch completing), the URI recalculates and the WebView scrolls back to the comment.

The root cause is that originalLinkingPath is stored in navigation params — it drives the WebView URI, but navigation params can be re-dispatched. Once the user has loaded the page, originalLinkingPath should be cleared from navigation state so it doesn't re-drive the URI on re-renders. The pattern used for hasLoadedUser.current (a ref that suppresses re-mounts) should be applied similarly to originalLinkingPath.

### **What was changed**
apps/backend/api/models/[Notification.js](http://notification.js/)
Added routeToPath(routeURL) helper at the top that returns pathname + search instead of pathname alone. Applied to every push sender in the file — comment, post, event invitation, contribution, track, donation, funding round, member join, and all group-relationship types.

sendCommentPush: moves the enabledNotification guard to the top (avoids loading the group for disabled users), then resolves the group via activity.get('group_id') first (the exact context row), falling back to groupForFrontendRouteForUser only when group_id is null or the group has been deleted. Handles public posts gracefully — group=null produces /public/post/id?commentId=xxx.

;sendPostPush: same group-resolution strategy — activity.get('group_id') first, then Activity.groupIds()[0] as fallback, throws only when neither is available.
apps/backend/api/services/[OneSignal.js](http://onesignal.js/)
Reads process.env.DOMAIN (the same env var that [Frontend.js](http://frontend.js/) uses) to build hyloapp://[www.hylo.com/path](http://www.hylo.com/path) on production or hyloapp://[staging.hylo.com/path](http://staging.hylo.com/path) on staging. Eliminates the hyloapp:/ + /path = hyloapp://groups-as-hostname corruption entirely.

apps/mobile/src/navigation/linking/[index.js](http://index.js/)
Added hyloapp://[www.hylo.com](http://www.hylo.com/), hyloapp://[staging.hylo.com](http://staging.hylo.com/), and hyloapp://[hylo.com](http://hylo.com/) to prefixes so React Navigation's string-based prefix stripping produces a clean /groups/slug/post/id path. The old hyloapp:// entry is kept as a fallback for any legacy notification URLs already in flight.

apps/mobile/src/hooks/[useOpenURL.js](http://useopenurl.js/)
Added isHyloappURL check: if linkingURL.protocol === 'hyloapp:' and the host is one of the known Hylo hosts, treat it as an internal URL. This lets openURL navigate directly via getStateFromPath rather than bouncing through Linking.openURL, which was the fragile path that caused dead ends.

apps/mobile/src/hooks/[useOpenInitialURL.js](http://useopeninitialurl.js/) — Removed the else { setInitialURL(null) } branch. Previously this cleared the stored URL if auth was still loading when the effect first ran, creating a race where a cold-start notification tap would have its URL silently discarded. Now the URL persists in the store until loading becomes false and the effect re-fires to process it.

apps/mobile/[index.js](http://index.js/) — Added OneSignal.Notifications.addEventListener('click', ...) handler. When a notification is tapped, it logs a Sentry breadcrumb and stores the launchURL in useLinkingStore. This gives useOpenInitialURL one canonical place to pick up the URL regardless of whether the OS Linking event or the click handler fires first. Both paths write the same URL so any double-dispatch is idempotent (same originalLinkingPath → same WebView URI → no reload).

apps/web/src/routes/PostDetail/Comments/Comment/Comment.jsx — Added hasScrolledToComment ref. The useEffect that fires scrollIntoView after 500ms is now guarded by this ref, so a given Comment instance only scrolls once even if selectedCommentId / [comment.id](http://comment.id/) deps re-evaluate during a comments refetch or parent state change. The ref resets naturally when the component unmounts (user navigates away), so coming back to the same post via a new notification correctly scrolls again.


apps/backend/api/models/Notification.js
```diff
@@ -183,7 +190,7 @@ module.exports = bookshelf.Model.extend({
     const locale = this.locale()
     return Group.find(groupIds[0])
       .then(group => {
-        const path = new URL(Frontend.Route.group(group)).pathname
```
tibetsprague, line 186: im actually not clear why we do this at all? why cant we just use what was returned by Frontend.Route.xxx?

apps/backend/api/models/Notification.js
```diff
@@ -237,40 +244,57 @@ module.exports = bookshelf.Model.extend({
...
-    if (isEmpty(groupIds)) throw new Error('no group ids in activity')
-    // TODO: include all groups in the notification?
-    return Group.find(groupIds[0])
-      .then(group => {
-        const path = new URL(Frontend.Route.post(post, group)).pathname
-        const alertText = PushNotification.textForPost(post, group, firstTag, version, locale)
-        return this.reader().sendPushNotification(alertText, path)
-      })
+
+    // Prefer the activity's group_id (the exact context where the notification was generated)
```
tibetsprague: actually this is where we should prefer the first group that the reader is a member of. That would finally fix this issue: https://github.com/Hylozoic/hylo/issues/731 and https://github.com/Hylozoic/hylo/issues/700

apps/backend/api/models/Notification.js
```diff
@@ -237,40 +244,57 @@ module.exports = bookshelf.Model.extend({
...
       return Promise.resolve()
     }
-    return this.reader().sendPushNotification(alertText, path)
+
+    // Prefer the activity's group_id (the exact context where the comment notification was generated).
+    // Fall back to the first group the reader is a member of, or null for public posts.
+    const activityGroupId = activity.get('group_id')
+    let group = activityGroupId ? await Group.find(activityGroupId) : null
+    if (!group) {
+      group = await post.groupForFrontendRouteForUser(reader.id)
```
tibetsprague: oh yeah we can just use post.groupForFrontendRouteForUser for that above too

tibetsprague (approved):
fingers crossed! one small change below

apps/backend/api/models/Notification.js
thomasgwatson: Frontend.Route.* functions are designed to produce full web URLs for use in emails and links. routeToPath is the adapter layer that converts those to the bare path the push system expects. The bug was that the adapter (new URL(...).pathname) was silently discarding query strings like ?commentId=; the new one (pathname + search) preserves them.

apps/backend/api/models/Notification.js
thomasgwatson: Actually this is a good prompt to clean this up

tibetsprague:
why does this require mobile first, the backend? does that mean all old mobile app versions will break?

thomasgwatson:
> why does this require mobile first, the backend? does that mean all old mobile app versions will break?

Ok... so there are trade-offs here;

Yes, these changes will degrade the experience of old version users.

There is a way to avoid that: instead of using the Custom scheme (hyloapp://) we use universal links (https://) 

This risks a different problem for all users then; Universal links (https://) require Apple/Google to verify that your domain "owns" the app — they do this by fetching a file from your server (apple-app-site-association / assetlinks.json). This verification can fail or be delayed, and when it does the OS just opens the URL in the browser instead of the app. Custom schemes are guaranteed — iOS/Android will always open the app for hyloapp:// because no browser claims that scheme.

We could switch to universal links for a time period, and send out more "update your hylo app" messages and then switch back to custom scheme links....

thomasgwatson:
> why does this require mobile first, the backend? does that mean all old mobile app versions will break?

Ahhh thats annoying I must not have hit send on my comment for this and now its gone...

In short;
Yes it will degrade old versions. 

The tension here is between the use of a custom scheme `hyloapp://` versus universal links `https://hylo.com`.

Universal links are slower to resolve, since the app then has to check our servers to confirm that it should open that link in the Hylo App, and there is a little risk that it will bounce and open the link in the browser.... but they would weather the transitions we are making to route linking/handling in a more back-wards compatible way.


So we could switch to universal links for the transition, tell people to update their apps, and then after a few months switch back to custom scheme links.

thomasgwatson:
Actually, reading more into it, it seems like custom scheme urls are increasingly less popular and are a bit of a hold over from before universal links could even be used. So I'm just going to switch it over to universal links

tibetsprague:
this ready to go?

thomasgwatson:
Yes, it'll just require monitoring and checking in prod, so please let me
know when you are able to deploy because I want to be looking at it
straight away

PR #1391, opened 2026-05-22
Try to fix constantly being logged out on ios

tibetsprague:
This has not been tested yet, but AI thinks this will fix it


thomasgwatson:
Ok, building in bitrise now.

thomasgwatson:
Ok, this build is available on bitrise. Hopefully you don't experience the bug now when you play with it . I did point the build at staging but perhaps it needs to be tested when the build points at production? what do you think?

PR #1396, opened 2026-05-28
Remove all .connectors from the web repo

tibetsprague:
More cleanup and modernization to simplify codebase. As part of this convert components to function components using hooks


thomasgwatson (approved):
lgtm

PR #1403, opened 2026-05-28
Fix the mobile app from logging out every time it is opened

tibetsprague:
I will see if this web only change fixes it before trying anything on the mobile code side

### DIFF
--- a/docs/mobile-ios-session-logout-investigation.md
+++ b/docs/mobile-ios-session-logout-investigation.md
@@ -1,0 +1,351 @@
+# Mobile iOS session loss on cold start — investigation notes
+
+This document explains why users on the Hylo **iOS mobile app** (TestFlight / production) were forced to log in again after every cold start, even though the session cookie appeared valid in Proxyman and existed in **Redis** immediately after login.
+
+It complements the older [mobile-auth-cookie-workflow.md](./mobile-auth-cookie-workflow.md), which describes cookie plumbing in general. This doc focuses on the **regression**, **root cause**, and **fixes**.
+
+---
+
+## Summary
+
+| Finding | Detail |
+|--------|--------|
+| **Not primarily** | Backend returning `me: null` for a valid `req.session.userId` on the same request |
+| **Not primarily** | AsyncStorage failing to persist the cookie (Proxyman showed the cookie on `api.hylo.com` after reopen) |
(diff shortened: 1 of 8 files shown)


thomasgwatson (approved):
Well, lets see if it helps

docs/mobile-ios-session-logout-investigation.md
```diff
@@ -0,0 +1,351 @@
+# Mobile iOS session loss on cold start — investigation notes
+
+This document explains why users on the Hylo **iOS mobile app** (TestFlight / production) were forced to log in again after every cold start, even though the session cookie appeared valid in Proxyman and existed in **Redis** immediately after login.
```
thomasgwatson, line 3: tbf, we also have that report of it happening to one Android user as well (assuming that its the same issue)

docs/mobile-ios-session-logout-investigation.md
```diff
@@ -0,0 +1,351 @@
...
+It complements the older [mobile-auth-cookie-workflow.md](./mobile-auth-cookie-workflow.md), which describes cookie plumbing in general. This doc focuses on the **regression**, **root cause**, and **fixes**.
+
+---
+
+## Summary
+
+| Finding | Detail |
+|--------|--------|
+| **Not primarily** | Backend returning `me: null` for a valid `req.session.userId` on the same request |
+| **Not primarily** | AsyncStorage failing to persist the cookie (Proxyman showed the cookie on `api.hylo.com` after reopen) |
```
thomasgwatson, line 14: thats interesting...

docs/mobile-ios-session-logout-investigation.md
```diff
@@ -0,0 +1,351 @@
...
+| **Primary cause** | The **web app inside the WebView** runs `checkLogin`, often gets `me: null` on the first request (cookie jar timing), then tells **native to `logout()`**, which **destroys the Redis session** |
```
thomasgwatson, line 15: Ok, so I guess this is the core issue. Yet another cold-start issue on mobile...

tibetsprague: yes. i thought removing the logout would fix it but it just makes the app hang. also been trying to ensure the cookie syncs before the first request, no luck so far

tibetsprague:
close in favor of other one

PR #1404, opened 2026-05-30
Fix mobile login 2

tibetsprague:
This one seems to actually work!?

Now testing whether we actually need all of this stuff

### DIFF
--- a/apps/mobile/src/util/session.js
+++ b/apps/mobile/src/util/session.js
@@ -2,7 +2,7 @@ import AsyncStorage from '@react-native-async-storage/async-storage'
 import { Platform } from 'react-native'
 import Config from 'react-native-config'
 import CookieManager from '@react-native-cookies/cookies'
-import { isNull, isUndefined, omitBy, reduce } from 'lodash'
+import { isNull, isUndefined, omitBy, reduce, mapValues } from 'lodash'
 
 const COOKIE_KEY = Config.SESSION_COOKIE_KEY || 'hylo_session_cookie'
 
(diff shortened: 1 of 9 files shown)


apps/mobile/src/util/session.js
```diff
@@ -62,6 +65,23 @@ export async function ensureWebViewCookies () {
...
+ * Does not use CookieManager.get to verify: iOS filtering can disagree with what
+ * WKWebView actually sends on fetch. A short delay lets setCookie completions settle.
+ *
+ * Returns { cookieStr } — null cookieStr means there is no native session at all.
+ */
+export async function prepareWebViewCookies () {
+  const cookieStr = await getSessionCookie()
+  if (!cookieStr) return { cookieStr: null }
+
+  await ensureWebViewCookies()
```
thomasgwatson, line 79: So, this happens...but doesn't impact the return value of the function?

apps/mobile/src/util/session.js
```diff
@@ -123,6 +150,34 @@ function invalidPair (v, k) {
     ['HttpOnly', 'Expires', 'Max-Age', 'Domain', 'Path', 'Version'].includes(k)
 }
 
+/**
+ * AsyncStorage merge bugs or RN fetch merging can produce "name=enc,name=dec" inside one
```
thomasgwatson, line 154: Was this discovered via logging the cookie value? Or is this a guess?

docs/mobile-ios-session-logout-investigation.md
```diff
@@ -0,0 +1,351 @@
...
+---
+
+## Fixes
+
+### 0. Sync WebView cookie jar before first web GraphQL (recommended)
+
+**Native (`apps/mobile/src/util/session.js`):**
+
+- `prepareWebViewCookies()` — writes AsyncStorage cookies into the WebKit jar via `CookieManager.set`, then **reads them back** with `CookieManager.get` until they match (retries with backoff).
+- `HyloWebView` does **not** mount the WebView until `prepareWebViewCookies()` completes.
```
thomasgwatson, line 238: Ok, i think this explains one of my questions above

apps/mobile/src/util/session.js
tibetsprague: using Proxyman on my phone to see all requests happening and looking at the cookies

thomasgwatson:
This one is still in draft. Lets merge yeah? or do you want to just do a release off the branch?

thomasgwatson:
Assuming we aren't going in this direction

PR #1410, opened 2026-06-01
More link fixes

### LINKED ISSUE #1408: Join link redirect going to wrong place
From the WTR: "We were supposed to launch today but weren't fully set up (because we had a crazy week) so we were just trying to get it together today but...then discovered...the join link redirects to https://www.hylo.com/groups/work-that-reconnects-network/chat/general This leads to the chat (which we're not using). This is the same issue I shared with you a couple of weeks ago about how the notifications point to the chat as well and each new post has the chat #. Going there drops people in a strange place (not the most recent posts / not our homepage) so it's confusing. Can these links (especially the join link) be redirected to point to our Home Page? We are pausing the launch until we hear back from you. Hopefully we can go ahead with the launch tomorrow.”

tibetsprague:
fixes https://github.com/Hylozoic/hylo/issues/1408

Going to join link while already a member now goes to correct home view. 

non-chat Post notifications now go to home view, not always to /chat/general

### DIFF
--- a/apps/backend/api/services/Frontend.js
+++ b/apps/backend/api/services/Frontend.js
@@ -144,6 +144,28 @@ module.exports = {
       return url(`/members/${getModelId(user)}`)
     },
 
+    /**
+     * Generates the frontend URL for viewing a post in context.
+     *
+     * Routing rules:
+     * 1. Funding-round submissions get their own dedicated URL.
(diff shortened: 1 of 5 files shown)


thomasgwatson (approved):
lgtm, along I worry its going to result in merge conflicts with paid content

PR #1412, opened 2026-06-04
Alt auth setup mobile

thomasgwatson:
Testing this out on staging now, via the staging-june-26 branch


apps/backend/scripts/spike-mint-tokens.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Throwaway spike: validates that OIDCTokens.mintTokensForUser works against the local
```
tibetsprague: so we can remove this?

tibetsprague (approved):

apps/backend/scripts/spike-mint-tokens.js
thomasgwatson: yee

PR #1417, opened 2026-06-10
Add different positioning for save action on mobile or tablet

thomasgwatson:
This is up on staging and I need to test it out there before merging


tibetsprague (approved):
did this get tested yet?

thomasgwatson:
No, I think its best to do it on a device and I haven't got to it yet. Just putting the save button somewhere that the keyboard won't cover up in the profile edit view

PR #1430, opened 2026-06-12
Clean up search query

### LINKED ISSUE #1428: Search is timing out periodically
Need to clear up search query

thomasgwatson:
Closes #1428 

I did some testing on my local db and we can do some on staging but the proof will be in the (prod) pudding

1. Restructure the query (highest impact)
Group access is now applied inside the FTS subquery, with public posts/comments included. Implemented in FullTextSearch.js.

2. Add indexes on search_index ID columns
Partial indexes on post_id, user_id, and comment_id. Migration + createView() update.


tibetsprague (approved):
looks fine to me. my agent did something a bit different :)

thomasgwatson:
OOo what?

PR #1436, opened 2026-06-15
1433 search performance focus

### LINKED ISSUE #1433: Search Improvements
- Error handling and no results graphic
- Performance improvements?

thomasgwatson:
For context: https://github.com/Hylozoic/hylo/discussions/1435

I have tested whether it works but whether it is more performant is unanswered. Of course, even if we remove the search stall that could happen every ten minutes, that just by itself would be good.

Closes #1433


apps/backend/api/services/FullTextSearch.js
```diff
@@ -1,15 +1,25 @@
-/* globals bookshelf */
+/* globals bookshelf, Group */
 import { compact, omit } from 'lodash'
 
 const tableName = 'search_index'
 const columnName = 'document'
 const defaultLang = 'english'
+const searchStatementTimeoutMs = Number(process.env.SEARCH_STATEMENT_TIMEOUT_MS) || 20000
+const recencyHalfLifeSeconds = 1209600 // 14 days
```
tibetsprague, line 8: I generally like constants to be all caps

apps/backend/api/services/FullTextSearch.js
```diff
@@ -42,37 +65,54 @@ const createView = (lang, knex) => {
...
       using gin(${columnName})`, knex))
     .then(() => raw(`create index idx_search_index_post_id on ${tableName} (post_id) where post_id is not null`, knex))
     .then(() => raw(`create index idx_search_index_user_id on ${tableName} (user_id) where user_id is not null`, knex))
     .then(() => raw(`create index idx_search_index_comment_id on ${tableName} (comment_id) where comment_id is not null`, knex))
 }
 
+const applyMemberGroupFilter = (subquery, groupAccess) => {
+  const { groupIds, userId } = groupAccess
+  if (userId) {
+    subquery.whereIn('group_id', Group.selectIdsForMember(userId))
```
tibetsprague, line 90: since Group.selectIdsForMember(userId) is being queried multiple times seems like we could query for it once ahead of time and then include that list here?

tibetsprague, line 90: though hopefully thats what postgres actually does on its own

apps/backend/lib/graphql-bookshelf-bridge/util/presentQuerySet.js
```diff
@@ -11,16 +11,18 @@ export default function presentQuerySet (models, options, querySetId) {
 
   var total = 0
 
-  if (options.total) {
+  if (options.total != null) {
     total = Number(options.total)
-  } else if (models.length > 0) {
+  } else if (options.hasMore != null) {
```
tibetsprague, line 16: this is a slightly concerning change as it it could effect something else that expects total to be null when using hasMore? probably ok but we should do some testing across the app

tibetsprague: do we even need this change? is total used somewhere?

thomasgwatson: Screenshot 2026-06-15 at 7 15 30 PM
Screenshot 2026-06-15 at 7 17 23 PM

These are the relevant parts of the overview

I planned this with a higher power model and executed with auto.

`total` shows up a lot in the pagination query sets but I don't think we directly use it... we should do a little more testing to see if search breaks without it

thomasgwatson:
@tibetsprague seems pretty crisp on staging...

How else do you want to test it out?

tibetsprague:
tested a bit more and looks good

tibetsprague (approved):

Issue #1135, opened 2025-12-09
Leaving Group Messages


AlchemicalSpiralizer:
All views and platforms.

Recently I was included in a message that was sent to many members of Hylo. The specific message was innocuous however it wasn't possible to leave or remove myself from the message group. Remaining part of the group means I get notifications every time any of the nearly 100 people in the message reply. This might result in forced persistence in the group despite having no interest in the context - it could lead to spam in the future.

MsNita:
<img width="1093" height="945" alt="Image" src="https://github.com/user-attachments/assets/5e78d721-fe84-46ba-905c-e20d76cb2c42" />

tibetsprague:
my latest thinking is we should have a three dot menu next to each thread with an option to Mute (and/or maybe Archive separately?), Unmute if muted, and one to fully leave with a warning that you can never get back again, and one to Mark as read/Unread

PR #1440, opened 2026-06-15
Restrict DMs to 20 participants and allow participants to leave a DM …

thomasgwatson:
…group

#1135


tibetsprague:
i think we decided to just implement mute for now which means you dont get notifications for those DMs anymore and we show a little mute icon next to it. Also probably dont show an orange dot next to? though maybe still show a dot indicating # of new messages and make it gray

thomasgwatson:
I've done muting, but where there are two lists; inbox and muted. So the muted threads are requested and displayed separately.

Issue #1375, opened 2026-05-08
Separate out Me/Auth


thomasgwatson:
#1358 prep work

thomasgwatson:
fixed

PR #1448, opened 2026-06-16
Improvements to mobile auth

tibetsprague:
Improve mobile WebView auth recovery UX with in-place retry and steady spinner. When the web app detected a cookie desync it remounted the entire WebView on each VERIFY_AUTH, causing visible reload flashes before logout.

Mobile:
- Re-mint the session on VERIFY_AUTH and post SESSION_READY so the web app can re-run checkLogin in place instead of reloading the document
- Sync session cookies to both HYLO_WEB_BASE_URL and API_HOST so GraphQL requests see the same session as native
- Keep PrimaryWebView’s spinner overlay up during recovery (AUTH_SUCCESS/LOGOUT clears it)

Web:
- Re-run checkLogin in place on SESSION_READY from native instead of requiring a document reload (pairs with an upcoming mobile release)
- Persist recovery state in localStorage so the fullscreen spinner survives WebView reloads on older app builds
- Always show fullscreen loading in HyloMobileV2 during initial load and cookie recovery (fixes blank /app on first open)
- Extract runCheckLogin for reuse on mount and SESSION_READY
- Add WebViewMessageTypes.SESSION_READY to shared constants

### DIFF
--- a/apps/mobile/src/components/HyloWebView/HyloWebView.js
+++ b/apps/mobile/src/components/HyloWebView/HyloWebView.js
@@ -4,7 +4,7 @@ import Config from 'react-native-config'
 import useRouteParams from 'hooks/useRouteParams'
 import AutoHeightWebView from 'react-native-autoheight-webview'
 import { getSessionCookie, clearSessionCookie, ensureWebViewCookies, sessionCookieFromToken } from 'util/session'
-import { parseWebViewMessage } from '.'
+import { parseWebViewMessage, sendMessageFromWebView } from '.'
 import { useAuth } from '@hylo/contexts/AuthContext'
 import { WebViewMessageTypes } from '@hylo/shared'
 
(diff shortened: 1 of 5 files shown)


thomasgwatson:
Building this in Bitrise

thomasgwatson:
Was noticing a lot of extra loading spinners between transitions (opening a post/chat detail view, closing it) on staging.

Reverted one of the branches I was working on; no change
Reverted this branch: the issue was gone. 

So I'm going to see if I can address on this branch

thomasgwatson:
Ok, I think I got this sorted.... effect dependencies arrays, such a consistent source of bugs in React :\

### Root cause:

The culprit is a dependency-array regression in the branch's RootRouter.js refactor.

The colleague extracted the session check into runCheckLogin and changed the mount effect's dependencies from [] to [runCheckLogin, navigate]. Because this app uses the non-data router, useNavigate() returns the unstable variant whose identity changes on every route change (it has locationPathname in its own deps). So:

Open the chat modal → URL changes → new navigate identity → the effect re-fires → runCheckLogin() → setLoading(true) → root spinner, then on resolve <AuthLayoutRouter/> remounts and shows its bootstrap skeleton before content.
Close the modal → URL changes again → same thing.
That's the spinner→skeleton flash on every in-app transition. On dev, this effect ran exactly once (mount), so navigations were instant with the chatroom staying rendered.

thomasgwatson:
I have rolled these changes into #1375 and #1376 and would suggest closing this one and figuring out things in 1376

PR #1463, opened 2026-06-25
Avoid double posting

### LINKED ISSUE #1461: Double-posting doable in mobile-webview
I was able to create duplicate posts in a stream context while creating a post. Just hit the send/post button twice real-fast

thomasgwatson:
Closes #1461

### DIFF
--- a/apps/web/src/components/PostEditor/PostEditor.js
+++ b/apps/web/src/components/PostEditor/PostEditor.js
@@ -273,7 +273,7 @@ function PostEditorInner ({
   /** Set to true when the post has been successfully submitted, preventing draft saves during teardown/navigation. */
   const isSubmittedRef = useRef(false)
   /** Blocks duplicate create/update dispatches before Redux pending state updates. */
-  const isSavingRef = useRef(false)
+  const isSubmittingRef = useRef(false)
 
   // Default topic to use when not in a chatroom — available immediately from the store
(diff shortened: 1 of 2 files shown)


apps/web/src/components/PostEditor/PostEditor.js
```diff
@@ -1134,116 +1139,130 @@ function PostEditorInner ({
...
-    const fileUrls =
-      fileAttachments && fileAttachments.map((attachment) => attachment.url)
-    const postLocation = currentPost.location || selectedLocation
-    const actualLocationId = await ensureLocationIdIfCoordinate({
-      fetchLocation,
-      postLocation,
-      locationId
-    })
+    if (isSavingRef.current) return
+    isSavingRef.current = true
```
tibetsprague: seems a bit odd that we need a isSaving ref and a isSubmitting state. Ai loves using refs, im a bit dubious about all of them but 🤷‍♂️

tibetsprague (approved):
seems like a whole lot of extra double checks but ok

apps/web/src/components/PostEditor/PostEditor.js
thomasgwatson: Yes, these could be stream-lined. The ref is preferred because its sync and thus blocking; otherwise useState being async means that there is still time for a user to double click. I've updated to remove the dup but sticking with the ref

PR #1464, opened 2026-06-25
Spaces and Views

tibetsprague:


docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+---
+
+### 2.2 `tracks` table — columns removed and added
+
+Tracks are now spaces with some special fields that are still on the tracks table. They also have a special view type `track-actions` that is used to display the actions of the track.
+
+```sql
+-- Add new FK pointing to the space created for this track
+ALTER TABLE tracks ADD COLUMN group_id bigint REFERENCES groups(id) ON DELETE SET NULL;
```
thomasgwatson, line 106: almost want to call this space_id to make it clearer, in lieu of future-us or agents seeing group_id and missing that its not the actual parent group

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+CREATE INDEX idx_tracks_group_id ON tracks(group_id);
+
+-- Remove columns migrated to the space group
+ALTER TABLE tracks
```
thomasgwatson: Is there a data migration that runs before this? This is just the plan I guess, we can break these changes up into separate migrations and do the clean-up of tables/columns on the follow-on migration

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+---
+
+### 2.5 `group_views` table — new table
+
+```sql
+CREATE TABLE group_views (
+  id               bigserial PRIMARY KEY,
+  group_id         bigint NOT NULL REFERENCES groups(id) ON DELETE CASCADE,
+  name             varchar,
```
thomasgwatson, line 164: https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don't_use_varchar(n)_by_default

If anything I default to `text`

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+| `map` | null | optional | — | — | — | — | — | — | — |
+| `welcome` | null | optional | — | — | — | ✓ | — | — | — |
+| `custom` | **required** | optional | optional | `{postTypes, activePostsOnly, defaultSort, defaultViewMode, searchText}` | — | — | — | — | — |
+| `collection` | **required** | optional | — | — | — | — | — | — | — |
+| `link` | **required** | optional | — | — | ✓ | — | — | — | — |
+| `post` | optional | optional | — | — | — | — | ✓ | — | — |
+| `member` | optional | optional | — | — | — | — | — | ✓ | — |
+| `space` | — | — | — | — | — | — | — | — | ✓ |
+| `group` | optional | optional | — | — | — | — | — | — | ✓ |
+| `text` | **required** | optional | — | — | — | optional | — | — | — |
```
thomasgwatson: header or heading might be a bit better

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+  "searchText": ""
+}
+```
+
+---
+
+### 2.6 `group_views_users` table — new table
+
+```sql
+CREATE TABLE group_views_users (
```
thomasgwatson, line 242: How are we handling toggling submenus open/closed? Do we want to store that for users in the db so its saved?

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+### 3.2 New model: `GroupViewUser`
+
+New file: `apps/backend/api/models/GroupViewUser.js`
+
+```javascript
+view() → belongsTo(GroupView, 'view_id')
+user() → belongsTo(User, 'user_id')
+```
+
+Static: `GroupViewUser.findOrCreate(viewId, userId)`, `GroupViewUser.markRead(viewId, userId)`, `GroupViewUser.incrementNewPostCount(viewId, userIds)`.
```
thomasgwatson: Do we take this as a chance to switch this to be like DM threads and keep track of the time of lastReadAt and then calculate the count based on that?

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+### 3.9 Queries that must exclude spaces
+
+Add `WHERE type != 'space'` or `WHERE parent_id IS NULL` wherever group lists are returned and spaces should not appear:
+
+| Query / Context | Change |
+|-----------------|--------|
+| Global nav groups list | Exclude spaces |
+| Related Groups view | Exclude spaces |
+| Group search / explore | Exclude spaces |
```
thomasgwatson: I think we'll need to check the generic search functionality there as well, to make sure how it handles groups makes sense

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+```graphql
+type GroupView {
+  id: ID
+  group: Group
+  name: String
+  type: String
+  order: Int
+  icon: String
+  pageContent: String
```
thomasgwatson, line 524: whats this one do?

thomasgwatson: ahh ok, this includes welcome etc

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+```
+
+Remove (Phase 6, after ContextWidget code removal):
+```graphql
+# contextWidgets, chatRooms, homeWidget
+```
+
+### 4.3 Remove types (Phase 6)
+
+Remove `CustomView`, `Collection`, `GroupWidget` (from GraphQL schema and all resolvers).
```
thomasgwatson: GroupWidget is the other one, not context widget related. We should probably still nuke it (and nuke farm groups ;P )

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+5. **Migrate members** from `tag_follows` (`tag_id = tag.id, group_id = parent_group_id`, settings.notifications set to some value, otherwise ignore this tag_follows and dont migrate it):
+   - Create `group_memberships`: `group_id = new_space.id`, `user_id`, copy settings from parent membership but set `postNotifications` from `tag_follow.settings`.
+   - Create `group_views_users`: `view_id = chat_view.id`, `user_id`, `new_post_count` and `last_read_post_id` from tag_follow.
+
+6. **Migrate posts:** chat posts in `groups_posts` for parent group that have this tag → reassociate with new space in `groups_posts`; remove parent group association.
+
+### Step 9 — Handle `#general` or other home chat rooms specifically
+
+- Remove `#general` from all posts: delete rows from `posts_tags` where `tag_id = #general.id`.
```
thomasgwatson: I feel like data migrations have struggled with just doing something like this without timing out before. Obviously it would be great to have all of these migrations carried out in one transaction but I fear it'll choke in a lot of circumstances...

I wouldn't be opposed to a dedicated data migration table, that can be updated during each step and the process can be resumed if interruptted AND/OR seeing if we can break some parts of this off as separate migrations (for example, I think we could do all the role table changes separate)

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+`groups.slug` remains globally unique for all groups including spaces.
+
+**Stored slug for spaces:** `{parentSlug}-{localName}` e.g. parent = `my-community`, space named "General" → stored slug = `my-community-general`.
+
+**Routing:** `/groups/my-community/spaces/general` — the URL uses only the local portion. The backend derives the stored slug as `parentSlug + '-' + spaceSlug`.
+
+**Collision handling:** if `my-community-general` already exists as a top-level group, append a number: `my-community-general-2`, `-3`, etc. Show this to the steward during space creation so they can adjust the name.
+
+**Promoting a space to a group:** The stored slug stays `my-community-general`. It now appears as `/groups/my-community-general` — a valid globally-unique group slug.
```
thomasgwatson, line 967: I wouldn't be mad if in this sort of future moment we allow a slug migration as well

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+**`FundingRoundHome.jsx`:** Currently has About/People/Submissions/Manage/Chat tabs. After:
+- About tab → `welcome` view (`order = 0`, home). Renders welcome page_content, and also round banner, phase status, timeline dates, voting method... below. **No join CTA** inside the space.
+- Submissions → `funding-round-submissions` view (`order = 1`). Non-deletable.
+- People tab → `members` view. This directory shows each member's **role in the round** — whether they can submit and/or vote, derived from `funding_round.submitter_roles` and `funding_round.voter_roles` cross-referenced with the member's group roles.
+- Chat tab → `chat` view.
+- The existing `AboutTab`, `SubmissionsTab`, `ManageTab` sub-components are repurposed as renderers within their respective `GroupView` types.
+- Manage tab → is now inside the space settings modal
+
+**Space join interstitial (new):** When a non-member navigates to any space, they see an interstitial page (not the full space content) showing:
```
thomasgwatson: not just the /about view for that space?

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+viewType='separator'                 → Not a route; menu divider only
+viewType='post'                      → Single post view
+viewType='member'                    → Member profile view
+viewType='group'                     → Navigate to the group
+```
+
+### 7.2 TrackHome and FundingRoundHome — remove tabs
+
+**`TrackHome.jsx`:** Currently has tabbed interface. After this change:
+- Tab UI removed entirely.
```
thomasgwatson: We'll have to UX test what feels best for mobile

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+- Simple drag-and-drop vertical list (no nested containers).
+- Per-row: settings gear which opens settings modal with different settings per view type. Custom view enables editing name, icon, settings for example. Also set as home and delete buttons.
+- Clicking settings icon next to a `welcome` type view → opens welcome page editor (replaces `WelcomePageTab`).
+- "Add View" → picker of available view types.
+- "Add Space" → space creation form.
+
+### 7.4 ContextMenu — always-visible bottom section
+
+**For all group members** (when applicable):
+- **Moderation** link → `/groups/:slug/moderation` — shown when `group.flaggedItemsCount > 0`
```
thomasgwatson: To clarify; once there is a single moderation action, this is permanent?

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+1. Name + slug (auto-generated, editable)
+2. Icon (Lucide picker)
+3. Description
+4. Accepted post types (pill toggles)
+5. Initial views (checkboxes for common view types) is displayed during creation, afterwards edited directly in the menu.
+6. **Access** — single selector setting both `groups.visibility` and `groups.accessibility`:
+   - **Open to all** — anyone in the group can see and join _(protected, open)_
+   - **Restricted** — anyone in the group can see it, but must request to join _(protected, restricted)_
+   - **Role-gated** — only members with a specific role can see and join _(protected, open; sets `required_roles`)_. Shows role picker when selected.
+   - **Paid** — anyone can see it, but must pay to join _(protected, restricted; sets `group.paywall = true`)_. Only shown if the parent group has paid content enabled.
```
thomasgwatson: Feature (maybe even a paid one): allow spaces to attach their own stripe account

Either way, it would be really good to surface group/space structure in the metadata of transactions related to a group, so that admins can track where funds are coming from

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+Space invitations surface in **My Invites** (the existing invites UI in the My context). No new UI needed — invite system works for spaces since they are groups. Ensure invite flows handle `parent_id IS NOT NULL` correctly.
+
+---
+
+## 8. Notifications & Unread Tracking
+
+### Per-view unread counting
+
+When a post is created in a group/space, increment `group_views_users.new_post_count` for matching views for all members:
```
thomasgwatson: As mentioned above somewhere; pros/cons of changing this to a timestamp, like DMs?

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+## 9. Search
+
+### Frontend
+
+Per-view search box (inherited from Stream) filters posts within the current view. No UI change needed.
+
+### Backend update
+
+Group-level search must now include child spaces. Update `searchQuery` to include:
+```sql
```
thomasgwatson: we'll need to do a permissions check on this as well

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+1. Existing flow continues.
+2. **Add step:** template selection. Template pre-configures `accepted_post_types` and initial views.
+3. Replace `Group.setupContextWidgets()` with `Group.setupGroupViews(groupId, template, trx)`.
+
+**Default views for all new groups (unless template overrides):**
+- `all` (All Activity), order=0 (home)
+- `chat`, order=1
+- `map`, order=2
+- `members`, order=3
```
thomasgwatson: disagree about default `map`, and `about` inclusion in the menu. Isn't map covered as a viewmode of all? If admins want a map menu item, then they can add one, no?

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+- Administration responsibility required to create a space in a group.
+- Via ContextMenu edit mode "Add Space" button.
+- `createSpace` mutation → creates `groups` row with `type = 'space'`, `parent_id`, then calls `Group.setupSpaceViews()`.
+- Automatically adds a `type = 'space'` `group_views` row in the parent group's menu.
+
+### Track/Funding Round space creation (after Phase 4)
+
+Creating a new Track or Funding Round automatically creates its space:
+- Both seed: `welcome` (order=0, home), specialized view (order=1, non-deletable), `chat` (order=2), `members` (order=3).
```
thomasgwatson: does it make sense to have specific routing rules for these two, in lieu of a 'home' view? For example, if you are enrolled in a track, it takes you to  the specialized view

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+- Chat activity cards in All Activity stream
+- Pinned posts per view
+- Kanban view mode
+- View renaming / aliases
+- Promote a Space to a Group (architecture supports it; no UI)
+- Analytics per space
+- Moderation queue scoped per space
+- Category system for posts
+- Spaces with locations on map
+- Project posts → Project Spaces migration
```
thomasgwatson, line 1553: Just thinking about complications here:

One thing we would lose here is that projects won't be able to include users that aren't a parent group member.

This is one of the edge cases where I could see a space belonging to more than one group, which raises a range of other issues

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+- "Add Space" / edit space UI in ContextMenu edit mode
+- Space creation form + mutations wired up
+- Space membership join/leave UI (preview for non-members)
+- Archive/unarchive space UI
+- Space Settings page
+- Space invites in My Invites
+
+### Phase 4 — Paid / Paywalled Spaces
+
+Generalizes the existing track paywall to work for any space. A group with paid content enabled can set any space to "Paid" access.
```
thomasgwatson: Might be a few speed bumps here: tracks can only be be one-time purchases at the moment, so there will still need to be some distinction between them and group paywall/offerings

docs/spaces-and-views-engineering-spec.md
```diff
@@ -0,0 +1,1284 @@
...
+
+**Backend:**
+- Existing Stripe offerings/products currently linked to `tracks.id` → change to link to `groups.id` (the space group)
+- "Add Track for an offering" flow → "Add Space for an offering"
+- `track.access_controlled` migration → set `group.paywall = true` on track spaces; migrate existing Stripe product association to the space group id
+- Paywall scope/access check switches from checking track access to checking space membership
+
+**Space join interstitial — paid CTA:**
+- When a non-member views a paywalled space, the interstitial shows the available offerings (same as the current `TrackPaywallOfferingsSection` content, now generalized)
+- Purchasing an offering creates a `group_memberships` row with paid status
```
thomasgwatson: This will need a tighter look at. Its possible to have a group_membership on a paywalled group without access, and access isn't tracked on group_membership itself, from recollection.

We'll need to look at the user_scope table is figure out what makes sense going forward and whether anything there needs a migration

thomasgwatson:
One thing I don't see explicitly mapped out is how parent group roles are injected into each of the spaces;
- assuming that admins have access to every space; do they automatically get added to them but with notifications turned off?
- Are admins actually added to spaces (inflates the member list, maybe in a weird way) or do they just have access to them (based on some Group method or something...). Either way, it will need to be reliably replicated across queries

And of course, when I say admin, we have to figure out what actual responsibilities the various rules apply to, since groups can make their own roles....

docs/spaces-and-views-engineering-spec.md
tibetsprague: i like that, although... what if we allow for track groups where the parent group is a track in and of itself?

docs/spaces-and-views-engineering-spec.md
tibetsprague: yeah the ordering of this is not exactly how it will go

docs/spaces-and-views-engineering-spec.md
tibetsprague: i guess so, but people could use this to just have a block of text too, though maybe that should be a different type that is styled differently in the menu

docs/spaces-and-views-engineering-spec.md
tibetsprague: as of now only spaces have sub-menus which are automatically toggled open when viewing the space and closed otherwise.

docs/spaces-and-views-engineering-spec.md
tibetsprague: last read id seems safer to me than time. with time you can run into edge cases around timing

docs/spaces-and-views-engineering-spec.md
tibetsprague: yeah just for welcome right now, but in the future we could allow for other pages which would be dope

docs/spaces-and-views-engineering-spec.md
tibetsprague: we probably just have to split into separate migrations. 
and yes the role changes i am doing on a separate branch first actually

docs/spaces-and-views-engineering-spec.md
tibetsprague: you mean allowing people to change their group slugs? totally, people ask us to do it for them all the time

docs/spaces-and-views-engineering-spec.md
tibetsprague: no i think this looks differently than about, but we'll see how the designs go

docs/spaces-and-views-engineering-spec.md
tibetsprague: im not sure, we should discuss

docs/spaces-and-views-engineering-spec.md
tibetsprague: for now lets just use the parent group stripe. definitely should track where they come from though

docs/spaces-and-views-engineering-spec.md
tibetsprague: new_post_count is not the important field, thats a cache to make the orange dots work, the important one is last_read_post_id which again is safer than a timestamp

docs/spaces-and-views-engineering-spec.md
tibetsprague: sure

docs/spaces-and-views-engineering-spec.md
tibetsprague: perhaps, or we just lock welcome as the home page? to discuss

docs/spaces-and-views-engineering-spec.md
tibetsprague: for the future

docs/spaces-and-views-engineering-spec.md
tibetsprague: gotcha

docs/spaces-and-views-engineering-spec.md
tibetsprague: sure thang

tibetsprague:
good points about space roles! we shall discuss

docs/spaces-and-views-engineering-spec.md
thomasgwatson: Yeah I think it'll be better as distinct things. While the header type won't have special backend qualities, I think it should have a UI feature that allows you to either move it individually or use it to move a cluster of everything between it and a separator. So we just have front-end logic that submits the ordering update as a block of updates, not just one. 

And then I don't think that text-block view would have that same behavior.

I guess we'll also have to revisit whether we want the About view to show text (purpose) or not

docs/spaces-and-views-engineering-spec.md
thomasgwatson: I don't quite follow your meaning here...

docs/spaces-and-views-engineering-spec.md
thomasgwatson: For lastRead, DMs seem to work happily, as do chats. I can't recall why it has tended to be a mess for groups overall/the stream? I suppose I assumed that the setup for DMs was just better and thus going for the timestamp option.

docs/spaces-and-views-engineering-spec.md
thomasgwatson: Yeah, a nice feature to have one day

docs/spaces-and-views-engineering-spec.md
thomasgwatson: Even if the action is resolved, it would still be good to be able to review/refer-back-to prior moderation stuff... so one way or another we'll need that. 

Assuming that the context menu for a group will still have a 'settings' cog, maybe shift it under there somehow.... allow all group members to see but not interact with group settings... Hmmm interesting transparency pondering

docs/spaces-and-views-engineering-spec.md
thomasgwatson: Yeah, I just imagine it'd be a little annoying if you are coming back to a track a few times and it takes you to the welcome page instead of the action next on your list

docs/spaces-and-views-engineering-spec.md
tibetsprague: well right now only spaces can be a track type and have the actions view, why not allow a parent group to be a track in and of itself? maybe there isnt a good use case for it but just a wondering

PR #1488, opened 2026-07-31
Improve sorting and filtering of member lists

### LINKED ISSUE #737: Members List - More Sorting Options
LONG TERM 

It would be great to have more sorting options for members. Maybe even something more interactive rather than a simple list. Something like what you can do in Airtable would be amazing. Example we did in the past - https://airtable.com/appRmzYWDdOIPqR0J/pagWnUBXQ2hCRPaAs

SHORT TERM

Add other sort options such as "Join Date" and "Role."

https://hylo.canny.io/admin/board/bugs-feature-requests/p/members-list-more-sorting-options

thomasgwatson:
Closes #737 

Screenshot 2026-07-31 at 2 32 00 PM


apps/web/src/routes/Members/Members.js
```diff
@@ -219,23 +273,58 @@ function Members (props) {
...
+function roleLabel (role) {
+  return `${role.emoji ? role.emoji + ' ' : ''}${role.name}`.trim()
+}
+
+function sortKeysFactory () {
+  return {
     name: 'Name',
-    location: 'Location'
+    location: 'Location',
+    join: 'Newest',
```
tibetsprague: let's call this Join Date, Newest isnt quite clear enough to me

tibetsprague (commented):
sweet! i do think we should probably add display of each person's join date and last activate date to the directory, at least when sorting by that, but probably always? @brodeur got an opinon here?

thomasgwatson:
Sweet, lets check in with Aaron et al about any tweaks and then I'll sort them out

tibetsprague (approved):

PR #1489, opened 2026-07-31
Filter deleted profiles out of queries on the backend

### LINKED ISSUE #874: Deleted Profiles are Displaying first in the member directory
Tibet says this bug is related to users joining with no name somehow?

Navigate to PHA and click on the member directory, you will see users with no name / profile image. They look like deleted profiles but they are stacking on the top creating a not ideal member directory experience! When I click on their profile, I see the axlotl. 

Reproduced this bug on Building Hylo.

https://hylo.canny.io/admin/board/bugs-feature-requests/p/deleted-profiles-are-displaying-first-in-the-member-directory

thomasgwatson:
closes #874


tibetsprague (commented):
so the real issue here is people somehow joining with an empty name. i was exploring how this is happening recently, something to do with partially complete signups. i dont think we should hide people with an empty name from the directory though... lets discuss in product

PR #1495, opened 2026-08-03
Added specific meetingLink to events/posts to improve events for onli…

### LINKED ISSUE #617: Allow zoom, jitsi, google meet links as event locations
Allow users to create online events by accepting an event link to whatever video conference service. That way, the user does not have to search in the event description for the link and the link will be a first class citizen in attached iCal files.

This may not be trivial if we don't want to add a new field to the post editor modal and instead reuse the location as either a geographical location or link to whatever video service.

thomasgwatson:
Closes #617 
Screenshot 2026-08-03 at 2 14 54 PM
Screenshot 2026-08-03 at 2 15 00 PM
Screenshot 2026-08-03 at 2 15 11 PM


tibetsprague (commented):
havent looked through the code yet, but i am wondering if instead of calling it meetingLink we create a more general URL field that could be used in other post types too? We can still label it meeting link for events. 

On the other hand, we could also just detect a Zoom link in the meeting text and display that as a join link on the event card/details. Same with Google Meet... This might be better? Let's discuss in product

thomasgwatson:
@tibetsprague I cut the auto-detect zoom/google meet/etc as a separate ticket. Based on our product discussion I think this is ready to merge as is.

tibetsprague (approved):

PR #1507, opened 2026-08-06
DRAFT: 506 recurring events

thomasgwatson:
For discussion


apps/backend/api/models/EventSeries.js
```diff
@@ -0,0 +1,144 @@
+import { GraphQLError } from 'graphql'
+import { pick } from 'lodash'
+import { RecurrenceHelpers } from '@hylo/shared'
+
+/*
+ * A recurring-event series. Each occurrence is a real event Post referencing
+ * this series via posts.event_series_id, so RSVPs, comments, ICS calendars and
+ * digests keep working per-occurrence with no special cases.
+ *
+ * Occurrences are materialized up front (no scheduled job): creation fills the
```
tibetsprague, line 10: why not generate more in a cron?

apps/backend/api/models/EventSeries.js
```diff
@@ -0,0 +1,144 @@
...
+    })
+  },
+
+  recurrenceRule: function () {
+    return this.get('recurrence_rule')
+  },
+
+  /**
+   * Creates event posts for not-yet-materialized occurrences, up to
+   * MAX_FUTURE_OCCURRENCES upcoming ones. Past occurrences don't count toward
```
tibetsprague, line 40: the one problem i could see is people looking a year into the future and not seeing the recurring event where its supposed to be and being confused by that. not quite sure how to fix. i wonder if in calendar view we make it fake posts on the front-end when people look into the future? and then if someone interacts with it we make it a real post?

docs/recurring-events-design.md
```diff
@@ -0,0 +1,105 @@
...
+
+posts (new columns)
+  event_series_id      bigint FK → event_series, indexed
+  original_start_time  timestamptz  (the slot this occurrence was generated for)
+```
+
+`original_start_time` serves two purposes:
+
+1. **Pattern anchoring.** Generation expands the rule from the series `start_time` and only creates occurrences after `max(original_start_time)` across *all* series posts (including deactivated ones) — so deleted occurrences are never recreated and a rescheduled occurrence never shifts the pattern.
+2. **Detached-occurrence detection (Phase 2).** `start_time != original_start_time` means the occurrence was individually rescheduled; series-wide edits should skip its time fields.
```
tibetsprague, line 48: if you reschedule one in the future, and then change the original events recurrence, im not actually sure what people would want in most cases 🤔

PR #1510, opened 2026-08-10
1506 locale fix date times

### LINKED ISSUE #1505: Allow en-gb / en-au locale to adjust date formats, and ensure other languages also change date format
For the subjects of the commonwealth and all non-americans world-wide

thomasgwatson:
Closes #1505 

This also completes a lot of the date/time parts of i18n that we never really tackled originally. And found some strings that were not translated when new features were added.


apps/backend/lib/util.js
```diff
@@ -26,23 +27,7 @@ module.exports = {
...
-      'fr-FR': 'fr-FR',
-      hi: 'hi-IN',
-      'hi-IN': 'hi-IN',
-      pt: 'pt-BR',
-      'pt-BR': 'pt-BR',
-      'pt-PT': 'pt-PT'
-    }
-
-    return lookup[locale] || 'en-US'
+    return normalizeLocaleToFull(locale)
```
tibetsprague: so then we can just get rid of mapLocaleToSendWithUS

apps/desktop/i18n.js
```diff
@@ -11,10 +15,16 @@ const initI18n = async () => {
...
+            return app.isPackaged ? `http://hylo.com/locales/${lng}.json` : `http://localhost:3000/locales/${lng}.json`
+          }
         },
-        fallbackLng: 'en',
-        supportedLngs: ['en', 'es'],
+        fallbackLng: {
+          [LOCALE_EN_GB]: [LOCALE_EN_US],
+          default: [LOCALE_EN_US]
+        },
+        supportedLngs: [LOCALE_EN_US, LOCALE_EN_GB, LOCALE_ES, 'en', 'es'],
```
tibetsprague: im confused. we support more languages than this. is this about what gets loaded initially? i think we should just load 'en' initially

apps/web/public/locales/fr.json
```diff
@@ -572,6 +572,7 @@
   "Ended: {{endTime}}": "Terminé : {{endTime}}",
   "Ends: {{endTime}}": "Se termine : {{endTime}}",
   "English": "Anglais",
+  "English (UK)": "Anglais (Royaume-Uni)",
```
tibetsprague, line 609: this is just about the date-time format yes? not other translations

apps/web/src/routes/AuthLayoutRouter/components/GlobalNav/LocaleDropdown/LocaleDropdown.js
```diff
@@ -26,34 +37,39 @@ export default function LocaleDropdown ({ renderToggleChildren, className }) {
...
           label: '🇪🇸 ' + t('Spanish'),
-          onClick: () => handleLocaleChange('es')
+          onClick: () => handleLocaleChange(LOCALE_ES)
         },
         {
-          key: 'de',
+          key: LOCALE_DE,
           label: '🇩🇪 ' + t('German'),
-          onClick: () => handleLocaleChange('de')
+          onClick: () => handleLocaleChange(LOCALE_DE)
```
tibetsprague, line 57: i dont quite understand why these constants are helpful

apps/web/src/store/models/index.js
```diff
@@ -95,6 +93,17 @@ orm.register(
   TopicFollow,
   Track,
   Widget
-)
+]
+
+/** Reuse one ORM instance in dev so Vite HMR does not re-register model descriptors. */
```
tibetsprague, line 98: huh?

apps/web/src/i18n.mjs
```diff
@@ -16,26 +38,48 @@ i18n
...
       caches: ['localStorage'],
       order: ['cookie', 'localStorage', 'sessionStorage', 'navigator', 'htmlTag'],
       lookupCookie: 'hylo-i18n-lng',
       lookupLocalStorage: 'hylo-i18n-lng',
-      lookupSessionStorage: 'hylo-i18n-lng'
+      lookupSessionStorage: 'hylo-i18n-lng',
+      convertDetectedLanguage: (lng) => normalizeLocaleToFull(lng)
+    },
+    fallbackLng: {
+      'en-GB': [LOCALE_EN_US],
```
tibetsprague, line 56: why do we need this? seems duplicative

packages/shared/src/DateTimeHelpers/DateTimeHelpers.ts
```diff
@@ -25,15 +26,9 @@ export interface FormatEventTimeDisplayResult {
...
-  switch (locale) {
-    case 'en':
-      return 'en-US'
-    case 'es':
-      return 'es'
-    default:
-      return 'en-US'
-  }
+/** Maps Hylo locale codes (short or full) to BCP 47 tags for Luxon/Intl date formatting. */
+export const getLocaleAsString = (locale?: string): string => {
```
tibetsprague: is this needed?

packages/shared/package.json
```diff
@@ -7,12 +7,12 @@
   "description": "Code shared across Hylo applications for consistency of results",
   "scripts": {
     "test": "jest",
-    "build": "echo 'Building shared package' && tsc && find dist/cjs -name '*.js' -exec sh -c 'mv \"$1\" \"${1%.js}.cjs\"' _ {} \\; && echo 'Shared package built'"
+    "build": "echo 'Building shared package' && tsc && echo 'Shared package built'"
```
tibetsprague: why this change?

packages/shared/package.json
```diff
@@ -7,12 +7,12 @@
...
   },
   "exports": {
     ".": {
       "import": "./src/index.js",
-      "require": "./dist/cjs/index.cjs"
+      "require": "./dist/cjs/index.js"
```
tibetsprague: why this change?

packages/shared/package.json
tibetsprague: node 24?

apps/backend/lib/util.js
thomasgwatson: Yeah, I'll do that

apps/web/public/locales/fr.json
thomasgwatson: Pretty much everyone in the world uses a different date format to the US. So there needs to be two englishes to represent that. Also, there were some missing translations and I added them  but I can't recall if it was on theis branch or another

apps/desktop/i18n.js
thomasgwatson: Yep this has needed to be updated since before these changes. I asked cursor to update en-gb, so it didn't necessarily see gaps in the other languages, which I think were originally added by Aaron?

apps/web/src/routes/AuthLayoutRouter/components/GlobalNav/LocaleDropdown/LocaleDropdown.js
thomasgwatson: its just to enforce using the same string literal across the code base.. We use this pattern a lot, why does it seem different here for you?

apps/web/src/store/models/index.js
thomasgwatson: This was an unrelated bug, surfaced by changing locales locally:

"The Spanish locale wasn’t the root cause — it was the trigger that surfaced a pre-existing Vite HMR + redux-orm dev bug.

What you saw
TypeError: Cannot redefine property: fromCollectionId
    at AliasedForwardsDescriptorInstaller.installForwardsDescriptor (redux-orm.js…)
The stack pointed at AttachmentManager in chat, but the failure was in redux-orm model setup, not date formatting itself.

What fromCollectionId is
In Collection.js, the Collection model has two many() relations:


Collection.js
Lines 26-27
  linkedPosts: many('CollectionPost'),
  posts: many('Post'),
When redux-orm registers models, it installs forward descriptors on related models — e.g. on Post it adds something like fromCollectionId for the reverse side of Collection.posts.

Those are installed with Object.defineProperty, which cannot be run twice on the same property.

Why it crashed when opening the date picker (in Spanish)
During Option B work we changed locale-related modules (util/locale.js, datetimepicker.jsx, calendar-util.tsx, @hylo/shared imports). In dev, Vite HMR re-evaluates changed modules.

That reload chain could re-import store/models/index.js, which previously did this on every module evaluation:

export const orm = new ORM({ stateSelector: state => state.orm })
orm.register(/* all models */)
On HMR, the same ORM instance could get register() called again. redux-orm then tried to re-install fromCollectionId on Post → crash.

Opening the date picker in Spanish happened to load those updated locale modules, so it looked like a Spanish bug. It would likely have happened with any locale change that triggered the same HMR path."

apps/web/src/i18n.mjs
thomasgwatson: It ensures that we don't need two sets of english translations; the en-gb locale will still use the original englsh. Of course, I could generate a whole other english translation file, with the correct (aka non-american ;P) spelling

packages/shared/src/DateTimeHelpers/DateTimeHelpers.ts
thomasgwatson: This PR attempts to stop us from generating two digit locale codes. But there could still be some floating around in the database. So the various functions are built to take either short or full locale strings and only return full ones.

packages/shared/package.json
thomasgwatson: Jeez, this is genuinely terrible. I suspect it missed the yarn build-packages command existed and tried to 'fix' the build.

Reverting this

apps/desktop/i18n.js
tibetsprague: I dont understand your response, why does supported languages show only a partial list?

apps/web/src/routes/AuthLayoutRouter/components/GlobalNav/LocaleDropdown/LocaleDropdown.js
tibetsprague: I never really liked this pattern, it's not like these strings are going to change, the constants do nothing. but im not going to make you change it

packages/shared/src/DateTimeHelpers/DateTimeHelpers.ts
tibetsprague: but all this function does is call another function

apps/desktop/i18n.js
thomasgwatson: Well that is quite weird because I have committed those changes, and confirmed they have been pushed. But I can't get it to render the correct file on github

Image

thomasgwatson: Ok, I suspect github has changed the default behavior or PRs after new commits are added... so you have to manually add more commits once your started a review 
Screenshot 2026-08-12 at 12 04 48 PM
Screenshot 2026-08-12 at 12 05 01 PM

tibetsprague (approved):

PR #1534, opened 2026-09-04
1188 hylo sandbox

thomasgwatson:
A demo, with synthetic data, available to non-logged in users.

Gives visitors the feeling of being a part of an active Hylo experience

available at /sandbox

Changes to the demo state are NOT saved between page loads. But the local mutations for most things should render


tibetsprague:
oops i merged staging into dev, instead of spaces and views! forgetting this was not yet really reviewed/tested

tibetsprague (approved):
looks good to me

thomasgwatson:
Ahh ok! I wasn't expecting it to go out straight away because I didn't want to slow down the release but I do think it'll be good to have it out in public to show off the updated experience. The tours also worked in it, last time I checked but lets poke it.

PR #1539, opened 2026-09-04
Zapier fixes

tibetsprague:
Allow for creating link previews and uploading image attachments via API calls like Zapier


thomasgwatson (approved):
lgtm

PR #1545, opened 2026-09-07
1543 mass grant access

### LINKED ISSUE #1543: Grant access: mass action
One group wants to have all existing members to have enduring access. Make it easier to grant access to things to more than one person at a time.

thomasgwatson:
closes #1543 

Screenshot 2026-09-07 at 5 10 56 PM

Screenshot 2026-09-07 at 4 51 30 PM


apps/backend/lib/grantCheckoutSessionAccess.js
```diff
@@ -70,7 +70,11 @@ async function grantCheckoutSessionAccess (session) {
       if (membership) {
         await membership.acceptAgreements()
       }
-      await GroupMembership.pinGroupToNav(userIdNum, accessGroupId)
+      // Spaces are not in GlobalNav; only pin the selling parent group
+      const accessGroup = await Group.find(accessGroupId)
+      if (accessGroup && accessGroup.get('type') !== 'space') {
+        await GroupMembership.pinGroupToNav(userIdNum, accessGroupId)
```
tibetsprague, line 76: why do we pin the group again? we just thought people would want that? im second guessing that

packages/graphql/queries/peopleQuery.js
```diff
@@ -15,6 +15,13 @@ export default gql`
             id
             name
             avatarUrl
+            memberships {
```
tibetsprague: where does this get used? loading memberships (and each group in each one) can be expensive

apps/backend/lib/grantCheckoutSessionAccess.js
thomasgwatson: If you paid for a group, it gets pinned. This is making sure that doesn't happen with spaces, since that doesn't make sense.

And yeah, its a guess at what "makes sense" but we can totally change it

packages/graphql/queries/peopleQuery.js
thomasgwatson: ahh this was to load members into the grant access people selector. Maybe we need light and heavy versions?

tibetsprague: you can always pass a variable in to the query to decide whether to load the memberships or not from where its called

PR #1548, opened 2026-09-09
Add Auto Add Members setting for spaces that adds every member in the…

tibetsprague:
… parent group to the space.

And new members who join the group.

People can still leave the space after the fact.


thomasgwatson (approved):
lgtm

PR #1549, opened 2026-09-09
Fix locale-dependent case ordering in i18n key sort

### LINKED ISSUE #1266: translation script sort reverses case
this line in /scripts/i18n/lib.js:

````
    // If case-insensitive equal, uppercase comes before lowercase
    return a.localeCompare(b, undefined, { sensitivity: 'case' })
````
does the opposite of the comment, lowercase is sorted before uppercase. It's a simple fix if it's truly an error

### PR DESCRIPTION
**Pull request by sandeep780049**
Fixes #1266

## The bug

`sortLocaleDeep` in `scripts/i18n/lib.js` documents its tie-break as *"If case-insensitive equal, uppercase comes before lowercase"*, but the implementation used:

```js
return b.localeCompare(a, undefined, { sensitivity: 'case' })
```

The reporter saw lowercase sorted before uppercase — contradicting both the comment and `lib.test.js`, which asserts `['A', 'a', 'B', 'b']`.

## Root cause

`localeCompare`'s ordering of case variants is **locale-dependent**. With the default runtime locale it happens to put uppercase first (so the test passed), but with other locales it does not — for example with Danish collation, `'a'.localeCompare('A', 'da', { sensitivity: 'case' })` returns `-1`, flipping the order. So the function contradicted its documented intent on machines with certain locales.

## The fix

Replace the tie-break with a locale-independent code-unit comparison (`a < b` / `a > b`), where uppercase code points are always smaller, so keys that are case-insensitively equal deterministically sort uppercase first on every machine — matching the comment and the existing test.

## Verification

- `npx jest scripts/i18n/lib.test.js` — 5/5 pass (including the uppercase-first assertion)
- Demonstrated the locale variance: `'a'` vs `'A'` with `sensitivity: 'case'` differs between `en`/`sv` (`-1`) and `da` (`+1`)

### DIFF
--- a/scripts/i18n/lib.js
+++ b/scripts/i18n/lib.js
@@ -26,8 +26,13 @@ function sortLocaleDeep (value) {
     if (lowerA !== lowerB) {
       return lowerA.localeCompare(lowerB)
     }
-    // If case-insensitive equal, uppercase comes before lowercase
-    return b.localeCompare(a, undefined, { sensitivity: 'case' })
+    // If case-insensitive equal, uppercase comes before lowercase.
+    // Use code-unit comparison (uppercase code points are smaller) so the
+    // order is deterministic and doesn't depend on the runtime locale:
+    // localeCompare's case ordering varies between locales (e.g. 'en' vs 'da').
+    if (a < b) return -1
+    if (a > b) return 1
+    return 0
   })) {
     sorted[key] = sortLocaleDeep(value[key])
   }


sandeep780049:
Rebased onto latest `dev` (`63c73a88`) — no conflicts. The E2E workflow run now shows as *action_required* (workflow approvals are needed for first-time contributors here), so a maintainer "Approve and run" click would be much appreciated 🙏

For context on the previous red ❌ run: the single failure (`members-pill-clamp.spec.js › members page pill clamp does not loop`) looks like an unrelated flaky test — I noticed the same test fail on E2E runs from other branches as well (e.g. run 34436255956). This PR only touches `scripts/i18n/lib.js` (i18n tooling), so it shouldn't affect web E2E behavior at all.

Happy to dig into that flaky pill-clamp test in a separate PR if that would be useful!

sandeep780049:
Small follow-up with good news: the rebase onto latest `dev` brought in b55c748f ("Fix e2e tests"), which adds a `test.setTimeout(120000)` to `members-pill-clamp.spec.js` — the exact test whose 30s timeout made the first CI run red (unrelated to this PR's change). So the pending E2E run on this branch should be green once approved. Thanks!

PR #1550, opened 2026-09-09
Add new page view type for static HTML content

tibetsprague:
Make is so smaller images dont full up the full width of the page.

Allow for uploading SVGs and imagex with spaces in the name.


apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/GroupViewCard.jsx
```diff
@@ -371,11 +371,6 @@ function GroupViewCard ({
   const lightSurfaceLabels = !isDark && !onPhoto
   const eventStart = eventStartForView(presented)
   const isSpace = presented.type === 'space'
-  const isWelcome = presented.type === 'welcome'
```
thomasgwatson, line 374: This is because "welcome" is covered by the new page type right?

thomasgwatson (approved):
nioce

apps/web/src/routes/AuthLayoutRouter/components/ContextMenu/GroupViewCard.jsx
tibetsprague: actually i removed extra content that was displaying on the card that i thought didnt look good

PR #1552, opened 2026-09-09
Use query param create=post instead of adding /create/post to any route to open create post modal

tibetsprague:
So we dont have to keep adding every new route and remembering to do that.

Also to simplify and unify with create group modal which is now create=group


thomasgwatson (approved):
lgtm

PR #1569, opened 2026-09-11
Add Remove Member action to profile page's three-dot menu

### LINKED ISSUE #332: Add Remove Member from Group to the three dot menu when looking at a user profile in a group as a person with Remove Users responsibility
When hitting the 3 buttons on the members profile, it only reads "Block Member", not "Remove"

https://hylo.canny.io/admin/board/bugs-feature-requests/p/moderator-removing-member-doesnt-work

jameswagner:
## Summary
- Members with the `Remove Members` responsibility could already remove someone from the group's Members list, but that same action was missing from an individual member's profile page — the three-dot menu there only offered "Edit Profile" (self) and "Block this Member" (others).
- Adds a "Remove member from group" option to that menu, gated behind the same `Remove Members` responsibility check used elsewhere (`RESP_REMOVE_MEMBERS` / `getResponsibilityTitlesForGroup`), reusing the existing `removeMember` GraphQL mutation and a confirmation dialog matching the pattern already used on the Members list (`components/Member/Member.js`).
- Hidden on your own profile and for the Axolotl system account, same as the existing Block action.
- No new translation strings needed — all i18n keys already exist in every locale file.

Closes #332

## Test plan
- [x] Added 2 unit tests to `MemberProfile.test.js` asserting the menu item appears when the current user has the `Remove Members` responsibility in the group, and is hidden when they don't
- [x] All 8 tests in `MemberProfile.test.js` pass
- [x] `standard` lint passes on the changed lines
- [x] Manually verified end-to-end against a local dev environment: logged in as the seeded Coordinator (`test@hylo.com`), opened another member's profile in a group, clicked "Remove member from group," confirmed, and verified in the database that their `group_memberships` row flipped to `active: false` and the group's active member count dropped accordingly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/web/src/routes/MemberProfile/MemberProfile.js
+++ b/apps/web/src/routes/MemberProfile/MemberProfile.js
@@ -1,5 +1,5 @@
 import { filter, isFunction } from 'lodash'
-import { Pencil, X } from 'lucide-react'
+import { Pencil, Trash2, X } from 'lucide-react'
 import React, { useState, useEffect, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import CopyToClipboard from 'react-copy-to-clipboard'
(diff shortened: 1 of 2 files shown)


thomasgwatson (approved):
lgtm

